PR #42, opened 2024-07-22
Supporting Tuples to superb lengths

### LINKED ISSUE #21: Implement up to 21 items in a tuple
https://github.com/tonybaloney/CSnakes/blob/bd04f9e00762bb0629e7e74bcf333c105c14227b/PythonSourceGenerator/Reflection/TypeReflection.cs#L46

```csharp

    private static TypeSyntax CreateTupleType(PythonTypeSpec[] tupleTypes)
    {
        var tupleTypeSyntax = new TypeSyntax[tupleTypes.Length];
        if (tupleTypes.Length > 8) // TODO: (track) Implement up to 21
        {
            throw new NotSupportedException("Maximum tuple items is 8");
        }

```

aaronpowell:
Moving to ValueTuple so we can use the `(string, string, ...)` syntax, as the compiler allows us to create that to any length and it will then generate the nested tuple stack for us.

Added a bunch of tests for up to 17 (three levels of nesting).

Fixes #21


tonybaloney (approved):

tonybaloney:
Tuple-tastic!

PR #63, opened 2024-08-02
Better environment initialisation

aaronpowell:
Making it so when you set the venv it will initialise that as well, assuming it doesn't exist (opt-out available via optional arg).

~~When the builder is building the `IPythonEnvironment` it will now try to install packages with pip, assuming a `requirements.txt` is found.~~ Rolled back this change and will tackle it in a new branch.

Some refactoring also.

- **Moving stuff around to be more C#-y**
- **Initializing the venv and providing a way to opt-out of doing so**
- ~~**Using pip to install deps if requirements.txt is found**~~
- **Moving the version of packages to props file for easier local dev**
- **Adding NuGet version to props file and enabling SourceLink for better debugging**
- **I think that adds the missing dependency**


src/CSnakes.Runtime/CSnakes.Runtime.csproj
```diff
@@ -6,6 +6,7 @@
 
     <ItemGroup>
         <PackageReference Include="pythonnet" Version="3.1.0-preview2024-07-03" />
+        <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All"/>
```
tonybaloney, line 9: What does this do?

src/CSnakes.Runtime/PythonEnvironmentBuilder.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            WorkingDirectory = home,
+            FileName = "pip",
+            Arguments = "install -r requirements.txt"
+        };
+
+        if (virtualEnvironmentLocation is not null)
+        {
+            string venvScriptPath = Path.Combine(virtualEnvironmentLocation, "Scripts");
+            startInfo.FileName = Path.Combine(venvScriptPath, "pip.exe");
```
tonybaloney, line 78: This is Windows specific

aaronpowell: ah good point, it shouldn't be `.exe` off Windows. Is the rest of the path more consistent?

tonybaloney: It's bin/pip for Nix and Scripts\pip.exe for Windows. But see latter comment about SoC

tonybaloney:
> > This is much cleaner, I wonder if we should move the ensure virtual environment code out though. It's going to be more complicated than pip install for a lot of setups, also for AI/ML projects Conda is likely to be the package manager just as much as pop.
> 
> 
> 
> Sure, we could have something that allows you to specify the package manager rather than it assuming pip. Or we could go down the path of making it more DI based, so you register a package manager like:
> 
> 
> 
> ```csharp
> 
> var builder = PythonEnvironmentBuilder.FromNuGet("3.12.3");
> 
> 
> 
> builder.Services.AddPip();
> 
> 
> 
> // and so on
> 
> ```
> 
> 

I think we'd inevitably want it to be DI/pluggable, if we want to implement this at all. Python devs are used to having to setup the virtual environment as a step before the code will work and the steps for doing so will vary between projects. 
I think that taking it out of this PR for now makes sense since it contains lots of other important changes then we can discuss the design of a package initialization tool.

src/CSnakes.Runtime/CSnakes.Runtime.csproj
aaronpowell: https://learn.microsoft.com/dotnet/standard/library-guidance/sourcelink

Makes debugging into external packages easier.

src/CSnakes.Runtime/PythonEnvironmentBuilder.cs
aaronpowell: Ah right. I'll pull this part out as a different PR.

aaronpowell:
Having to provision the venv makes sense at least, and it simplifies the CI environment.

I also think it'll be good to have package install as part of the builder pipeline, as it means you don't have to run additional stuff outside yourself before the .NET process starts up. But I'll pull that stuff out of this PR.

tonybaloney (approved):

PR #64, opened 2024-08-03
Setup a documentation site

tonybaloney:


tonybaloney:
The API is still in flux so this is mostly scaffolding

PR #69, opened 2024-08-06
Add comment about pytorch to phi3 demo

tonybaloney:


tonybaloney:
@aaronpowell  you can run the phi3 demo now by just doing `python phi3_demo.py`

PR #103, opened 2024-08-14
Add an example of invoking CSnakes directly

### LINKED ISSUE #102: Document how to use the runtime without the generator
This is totally possible, just needs to be documented.

tonybaloney:
Fixes #102


docs/advanced.md
```diff
@@ -33,3 +33,73 @@ env = app.Services.GetRequiredService<IPythonEnvironment>();
...
+```
+
+The C# code to call this function needs to:
+
+1. Use the TypeConverter to convert the .NET types to `PyObject` instances and back.
+1. Use the `GIL.Acquire()` method to acquire the Global Interpreter Lock for all conversions and calls to Python.
+1. Use the `Import.ImportModule` method to import the module and store a reference once so that it can be used multiple times.
+1. Dispose the module when it is no longer needed.
+
+```csharp
```
tonybaloney, line 55: @aaronpowell  can you have a look at this sample. It's a cut-down version of the generated code.

PR #132, opened 2024-08-19
remove path locator

aaronpowell:
- **Putting samples into CI build**
- **Making build log a bit more readable**
- **Deleting PathLocator as it's rather problematic**


tonybaloney (approved):

tonybaloney:
It needs removing from the samples as well

Issue #99, opened 2024-08-13
BigInteger marshalling as an option instead of `long`


tonybaloney:
The Python `int` type is an array of digits, not a fixed-size integer.

Currently, we're marshalling `PyLong` as a `long` in .NET which is a 64-bit integer `(long long)`.

There are scenarios where developers would handle Python integers beyond the overflow point and want to bring them into C#. 

C# has the [`BigInteger` struct](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.biginteger?view=net-8.0) which is a better candidate for conversion.

I'd like the option to use BigInteger, but don't want to make it the default.

tonybaloney:
This can be implemented in the converter instead of inside the type reflections, that way the user has the ability to handle it themselves.

PR #136, opened 2024-08-19
Allow conversion between Python `int` and `System.Numerics.BigInteger`

tonybaloney:
I also refactored the Converter into separate files as it's getting very hard to read. 

Implements #99


docs/advanced.md
```diff
@@ -1,5 +1,32 @@
...
+BigInteger input = BigInteger.Parse(number);
+
+using (GIL.Acquire())
+{
+    using PyObject? pyObj = td.ConvertFrom(input) as PyObject;
+
+    // Do stuff with the integer object
+
+    // Convert back
+    BigInteger integer = (BigInteger) td.ConvertTo(pyObj, typeof(BigInteger))!;
```
aaronpowell, line 26: We should make a method on `PyObject` that wraps the converter so people don't have to access it themselves, maybe have `public static T? PyObject.Create<T>(T value) => td.ConvertFrom(input);` and then another method on `PyObject` that is an instance method to do the `ConvertTo` (or is that available with the `As<T>` method?).

This way we can completely hide away the converter part and if we want to change it in the future we're not going to be breaking people.

docs/reference.md
```diff
@@ -160,6 +160,24 @@ All public methods generate by CSnakes have a built-in GIL acquisition and relea
 
 ## Exceptions
 
-CSnakes will raise a `PythonException` if an error occurs during the execution of the Python code. The `PythonException` class contains the error message from the Python interpreter.
+CSnakes will raise a `PythonInvocationException` if an error occurs during the execution of the Python code. The `PythonInvocationException` class contains the error message from the Python interpreter as the `InnerException`.
```
aaronpowell, line 163: That's not right is it? Isn't it `PythonRuntimeException` is the raised exception with an `InnerException` of `PythonInvocationException`

aaronpowell (changes requested):

docs/reference.md
tonybaloney: It's correct.

PR #137, opened 2024-08-19
Marshal `SafeHandle` impls through the P/Invoke

AaronRobinsonMSFT:
Calling `DangerousGetHandle` without calling the corresponding
`DangerousAddRef` and `DangerousRelease` is unsafe.

This PR uses the provided marshallers in the P/Invoke signatures
or manually marshals when passing an array of `SafeHandle`s.

New APIs added with the "Raw" suffix for use when the raw
handle is returned immediately from a P/Invoke.


aaronpowell:
Looks like some merge conflicts @AaronRobinsonMSFT before I can trigger the build - can you tackle them?

tonybaloney (approved):

PR #141, opened 2024-08-20
Implement Generator Iterators

### LINKED ISSUE #140: Functions which return PyObject are disposed inside the scope of the function
When the function returns the result object without conversion, it incorrectly uses a `using` statement and disposes of the result on return, causing all code trying to read the fields to break.

```csharp
public PyObject ExampleGenerator(long length)
{
    using (GIL.Acquire())
    {
        logger.LogInformation("Invoking Python function: {FunctionName}", "example_generator");
        using PyObject __underlyingPythonFunc = this.module.GetAttr("example_generator");
        using PyObject length_pyObject = PyObject.From(length);
        using PyObject __result_pyObject = __underlyingPythonFunc.Call(length_pyObject);
        return __result_pyObject;
    }
}
```

tonybaloney:
Also 
Fixes #140


src/CSnakes.Runtime/Python/GeneratorIterator.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    object IEnumerator.Current => Current!;
+
+    public void Dispose()
+    {
+        generator.Dispose();
+        nextPyFunction.Dispose();
+        using PyObject close = closePyFunction.Call();
+        closePyFunction.Dispose();
+        sendPyFunction.Dispose();
```
AaronRobinsonMSFT, line 31: Is the order here important? If not, I would write `closePyFunction.Call().Dispose()`. If it is, perhaps a comment on why?

src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -279,12 +279,23 @@ public override string ToString()
         }
     }
 
-    public T As<T>() => (T)(td.ConvertTo(this, typeof(T)) ?? default!);
+    public T As<T>()
+    {
+        using (GIL.Acquire())
+        {
+            return (T)(td.ConvertTo(this, typeof(T)) ?? default!);
```
AaronRobinsonMSFT, line 286: Instead of `default!`, I would update this to be `T?`. It is annoying in almost all cases, but it is clearer than someone getting a `null` when it doesn't indicate that. Alternatively, one could throw if the conversation fails? The name of the function does seem to indicate that `T?` is likely more appropriate though.

src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -279,12 +279,23 @@ public override string ToString()
...
-    public static PyObject? From<T>(T value) => 
-        value is null ?
-            new PyObject(CPythonAPI.GetNone()) :
-            (PyObject?)td.ConvertFrom(value);
+    public static PyObject? From<T>(T value)
+    {
+        using (GIL.Acquire())
+        {
+            return value is null ?
+                new PyObject(CPythonAPI.GetNone()) :
```
AaronRobinsonMSFT, line 295: `new PyObject(CPythonAPI.GetNone())` seems like it could be a common type to pass around. Perhaps subclass `PyObject` and override the implementation so it never frees anything? Then you can have a `static` version for all?

```csharp
class PyObject ...
{
    internal static PyObject None => new PyObjectNone();
    ...
}
```

src/CSnakes.Runtime/PyObjectTypeConverter.Generator.cs
```diff
@@ -0,0 +1,18 @@
...
+internal partial class PyObjectTypeConverter
+{
+    private object? ConvertToGeneratorIterator(PyObject pyObject, Type destinationType, ITypeDescriptorContext? context, CultureInfo? culture)
+    {
+        Type item1Type = destinationType.GetGenericArguments()[0];
+        Type item2Type = destinationType.GetGenericArguments()[1];
+        Type item3Type = destinationType.GetGenericArguments()[2];
+        Type generatorType = typeof(GeneratorIterator<,,>).MakeGenericType(item1Type, item2Type, item3Type);
+        ConstructorInfo ctor = generatorType.GetConstructors().First();
+        return ctor.Invoke([pyObject.Clone()]);
```
AaronRobinsonMSFT, line 16: This is very slow in practice. Any chance you can make this strongly typed? I assume that you can get the actual types from the Python script?

src/CSnakes.Runtime/Python/PyObject.cs
tonybaloney: I want to add a method `IsNone()` to PyObject as that's a common scenario, will include that in the PR. True and False are also Immortal objects (depending on the version of Python how that works)

src/CSnakes.Runtime/Python/PyObject.cs
tonybaloney: If the conversion fails, there is always an exception thrown so the response type isn't nullable anyway. We only have to do this because of the `TypeConverter` API. Will add to the list of things to fix when that is refactored.

src/CSnakes.Runtime/PyObjectTypeConverter.Generator.cs
tonybaloney: Will address in TypeConverter refactor :+1:

aaronpowell (changes requested):

src/CSnakes.Runtime/Python/GeneratorIterator.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        try
+        {
+            using PyObject result = nextPyFunction.Call();
+            current = result.As<TYield>();
+            return true;
+        }
+        catch (PythonInvocationException pyO) when (pyO.PythonExceptionType == "StopIteration")
+        {
+            return true
```
aaronpowell, line 48:
```suggestion
            return true;

```

My bad

src/CSnakes.Runtime/Python/GeneratorIterator.cs
```diff
@@ -0,0 +1,74 @@
...
+        try
+        {
+            using PyObject sendValue = PyObject.From<TSend>(value) !;
+            using PyObject result = sendPyFunction.Call(sendValue);
+            current = result.As<TYield>();
+            return current;
+        }
+        catch (PythonInvocationException pyO) when (pyO.PythonExceptionType == "StopIteration")
+        {
+            return true
```
aaronpowell, line 67:
```suggestion
            return true;

```

My bad

src/CSnakes.Runtime/Python/GeneratorIterator.cs
```diff
@@ -0,0 +1,76 @@
...
+        try
+        {
+            using PyObject sendValue = PyObject.From<TSend>(value) !;
+            using PyObject result = sendPyFunction.Call(sendValue);
+            current = result.As<TYield>();
+            return current;
+        }
+        catch (PythonInvocationException pyO) when (pyO.PythonExceptionType == "StopIteration")
+        {
+            return true;
```
aaronpowell, line 68:
```suggestion
            throw new ArgumentOutOfRangeException("Generator is exhausted");

```

My double bad

PR #148, opened 2024-08-21
Add IsNone and None static attribute for PyObject

tonybaloney:
Should reduce the number of object allocations on the C# side and also this is a more convenient way to get a None value to pass to Python.


tonybaloney:
Cc @AaronRobinsonMSFT I think this is what you were suggesting. Can you have a look at the Immortal object implementation please.

src/CSnakes.Runtime/Python/Interns/ImmortalPyObject.cs
```diff
@@ -0,0 +1,15 @@
+﻿using CSnakes.Runtime.CPython;
+namespace CSnakes.Runtime.Python.Interns;
+
+internal class ImmortalPyObject(nint handle) : PyObject(handle)
+{
+    protected override bool ReleaseHandle()
+    {
+        return true;
+    }
```
aaronpowell, line 9:
```suggestion
    protected override bool ReleaseHandle() => true;

```

src/CSnakes.Runtime/Python/Interns/PyNoneObject.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public override string GetRepr()
+    {
+        return "None";
+    }
+
+    public override string ToString()
+    {
+        return "None";
+    }
```
aaronpowell, line 24:
```suggestion
    public override bool IsNone() => true;

    public override string GetRepr() => ToString();

    public override string ToString() => "None";

```

src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -157,6 +158,25 @@ public string GetRepr()
...
     }
 
+    /// <summary>
+    /// Is the Python object None?
+    /// </summary>
+    /// <returns>true if None, else false</returns>
+    public virtual bool IsNone()
+    {
+        return CPythonAPI.IsNone(this);
+    }
```
aaronpowell, line 168:
```suggestion
    public virtual bool IsNone() => CPythonAPI.IsNone(this);

```

src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -157,6 +158,25 @@ public string GetRepr()
...
+
+    public static PyObject None
+    {
+        get
+        {
+            if (none is null)
+                throw new InvalidOperationException("Python is not initialized. You cannot call this method outside of a Python Environment context.");
+            return none;
+        }
+    }
```
aaronpowell, line 178:
```suggestion
    public static PyObject None => none ??
                throw new InvalidOperationException("Python is not initialized. You cannot call this method outside of a Python Environment context.");

```

I **think** that is right.

aaronpowell (changes requested):

src/CSnakes.Runtime/Python/Interns/PyNoneObject.cs
```diff
@@ -0,0 +1,25 @@
+﻿using CSnakes.Runtime.CPython;
+
+namespace CSnakes.Runtime.Python.Interns;
+
+internal class PyNoneObject : ImmortalPyObject
```
AaronRobinsonMSFT, line 5:
```suggestion
internal sealed class PyNoneObject : ImmortalPyObject

```

src/CSnakes.Runtime/Python/PyObject.cs
AaronRobinsonMSFT, line 178:
```suggestion
    public static PyObject None { get; } = new PyNoneObject();
```

This will avoid allow you to remove the `internal` field and avoid the `PyObject.none =` logic above.

You could then move the Python runtime initialization check to be in the `ImmortalPyObject` constructor?

AaronRobinsonMSFT (commented):
Yep, this is precisely what I was thinking.

aaronpowell (approved):

PR #151, opened 2024-08-21
Add methods to `PyObject` for `Is`, `Equals` and `NotEquals`

tonybaloney:
Also adds operator overloads. 

This now works:

```csharp
// Test rich comparisons of mixed types
var obj11 = PyObject.From(3.0);
var obj12 = PyObject.From(3);
Assert.True(obj11!.Equals(obj12!));
Assert.True(obj11 == obj12);
```

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs
@@ -82,5 +82,51 @@ public class PyObjectTests : RuntimeTestBase
         var obj5 = PyObject.From(42);
         var obj6 = PyObject.From(43);
         Assert.False(obj5!.Equals(obj6!));
+
+        var obj7 = PyObject.From(42123434);
+        var obj8 = PyObject.From(42123434);
+        Assert.True(obj7 == obj8);
+
+        var obj9 = PyObject.From("Hello");
+        var obj10 = PyObject.From<IEnumerable<string>>(["Hello"]);
+        Assert.False(obj9!.Equals(obj10!));
+
+        // Test rich comparisons of mixed types
+        var obj11 = PyObject.From(3.0);
+        var obj12 = PyObject.From(3);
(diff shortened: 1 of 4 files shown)


src/CSnakes.Runtime/CPython/Object.cs
```diff
@@ -154,4 +162,30 @@ internal static bool HasAttr(PyObject ob, string name)
...
+        int result = PyObject_RichCompareBool(ob1, ob2, Py_NE);
+        if (result == -1)
+        {
+            PyObject.ThrowPythonExceptionAsClrException();
+        }
+        return result == 1;
+    }
+
+    [LibraryImport(PythonLibraryName)]
+    internal static partial int PyObject_RichCompareBool(PyObject ob1, PyObject ob2, int opid);
```
AaronRobinsonMSFT, line 190:
```suggestion
    internal static partial int PyObject_RichCompareBool(PyObject ob1, PyObject ob2, [MarshalAs(UnmanagedType.I4)] CompareType opid);
```

You can then turn the list of constants into an `enum`.

src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs
```diff
@@ -1,69 +1,132 @@
...
-            using PyObject? pyObj = td.ConvertFromString("hello") as PyObject;
+            using PyObject? pyObj = PyObject.From("hello");
             Assert.NotNull(pyObj);
             pyObj.Dispose();
             Assert.Throws<ObjectDisposedException>(() => pyObj!.ToString());
         }
     }
+
+    [Fact]
+    public void TestObjectIs()
```
aaronpowell, line 57: Can we split these as separate test cases, it makes it easier to find when a test has failed which scenario it was rather than having to read the stack trace

aaronpowell (changes requested):

PR #156, opened 2024-08-22
Implement <, >, <=, >=

### LINKED ISSUE #153: Add operator overloads to `PyObject` for <, >, <= and >=
See the `==` and `!=` implementations in PyObject for how it works.

The `<=` and `>=` operators should do a fast-check against `Is` before doing a rich comparison.

AaronRobinsonMSFT:
Fixes #153 

Not sure how to write a test for this. Since the source generator mostly converts `PyObject`s to natural .NET types, I don't know where to write the test and am very ignorant of how to create an "object" in Python that forces the generator to not return a `long` for the following signature: `def int_one() -> int:`.

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs
@@ -3,7 +3,6 @@
 namespace CSnakes.Runtime.Tests.Python;
 public class PyObjectTests : RuntimeTestBase
 {
-
     [Fact]
     public void TestToString()
     {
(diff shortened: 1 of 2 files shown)


tonybaloney (approved):
Nice. this is really clean. 
The tests for Equals were written as an xUnit theory and we just had some inline data of things which are "equal" in Python. https://github.com/tonybaloney/CSnakes/blob/main/src/CSnakes.Runtime.Tests/Python/PyObjectTests.cs#L72-L84

AaronRobinsonMSFT:
@tonybaloney I added some tests and they are only failing on macOS with some very odd behavior. I will need to debug them.

aaronpowell (approved):
No comments other than the pattern matching is 🔥

PR #159, opened 2024-08-23
Improve performance by 2x and add a basic benchmark test

tonybaloney:
This pull request includes several changes aimed at improving the codebase by simplifying methods, optimizing performance, and enhancing readability. The most important changes include refactoring type conversion methods, modifying internal API functions, and optimizing dictionary and list handling.

### Refactoring type conversion methods:

* [`src/CSnakes.Runtime/PyObjectTypeConverter.BigInteger.cs`](diffhunk://#diff-4371cbdc8f3e972c8ccdcf8b63814c316eb4f5ee505a5915327bf4ad0a5492d4L3-R12): Simplified the `ConvertToBigInteger` and `ConvertFromBigInteger` methods by removing unnecessary parameters and directly returning the parsed `BigInteger`.
* [`src/CSnakes.Runtime/PyObjectTypeConverter.Dictionary.cs`](diffhunk://#diff-8753067858ef9c42230ceb0697070cb681121247fda084da6d2d3b0b5935db3eL5-R84): Refactored the `ConvertToDictionary` and `ConvertFromDictionary` methods to optimize key handling and reduce heap allocations.
* [`src/CSnakes.Runtime/PyObjectTypeConverter.List.cs`](diffhunk://#diff-2882a910b6d2423be7abe8ebaa4be56e7884078d3cf077d9c73a7f9e34d73f42L4-R63): Improved the `ConvertToList` and `ConvertToListFromSequence` methods by pre-allocating list sizes and simplifying the conversion process.
* [`src/CSnakes.Runtime/PyObjectTypeConverter.Tuple.cs`](diffhunk://#diff-33728fdfde5137e166b2334ed5b6c6c10bb4313e5e59241e62999c11538e17daL3-R73): Refactored the `ConvertFromTuple` and `ConvertToTuple` methods to streamline tuple handling and optimize nested tuple processing.

### Modifying internal API functions:

* [`src/CSnakes.Runtime/CPython/Bool.cs`](diffhunk://#diff-aeb2790935371282e306ce420ac6aa49e3cca161f87d067b5afd5942f2b635d2L22-R22): Updated the `IsPyBool` method to directly compare the handle with `Py_True` and `Py_False` for better performance.
* [`src/CSnakes.Runtime/CPython/Exceptions.cs`](diffhunk://#diff-eba6fc888aa9da7c4b3519803a5ae273407e3b46f42a348aef73ada74864a611L12-R23): Refactored the `PyErr_Occurred` method to return a boolean and introduced a private `PyErr_Occurred_` method for internal use.
* [`src/CSnakes.Runtime/CPython/Long.cs`](diffhunk://#diff-9602407e39a1aba18e9df2d2cfc94572fb21bccc741498c9243db86e3ec63a2dR19-R21): Added a new method `PyLong_AsLong` to handle long integer conversions.
* [`src/CSnakes.Runtime/CPython/Tuple.cs`](diffhunk://#diff-07b977cba86dee16de04bc3dad5e20390094a20b8de0cdc06c7bd974860e3527L70-R72): Introduced new methods `PyTuple_GetItemWithNewRef` and `PyTuple_GetItemWithNewRefRaw` to handle tuple item retrieval with new references. [[1]](diffhunk://#diff-07b977cba86dee16de04bc3dad5e20390094a20b8de0cdc06c7bd974860e3527L70-R72) [[2]](diffhunk://#diff-07b977cba86dee16de04bc3dad5e20390094a20b8de0cdc06c7bd974860e3527R81-R96)
* [`src/CSnakes.Runtime/CPython/Unicode.cs`](diffhunk://#diff-464c629312a2b08a39163cc21fd800530aace494ceba9fff440dd948bc6be7d0R20-R31): Added a new method `PyUnicode_AsUTF8Raw` for UTF-8 string conversion.

### Optimizing dictionary and list handling:

* [`src/CSnakes.Runtime/PyObjectTypeConverter.Dictionary.cs`](diffhunk://#diff-8753067858ef9c42230ceb0697070cb681121247fda084da6d2d3b0b5935db3eL5-R84): Enhanced dictionary conversion by directly working with raw handles and optimizing key-value pair processing.
* [`src/CSnakes.Runtime/PyObjectTypeConverter.List.cs`](diffhunk://#diff-2882a910b6d2423be7abe8ebaa4be56e7884078d3cf077d9c73a7f9e34d73f42L4-R63): Improved list conversion by pre-allocating list sizes and simplifying item conversion.

These changes collectively improve the efficiency and readability of the codebase, making it easier to maintain and extend in the future.


src/CSnakes.Runtime/PyObjectTypeConverter.Tuple.cs
```diff
@@ -31,7 +31,7 @@ private PyObject ConvertFromTuple(ITypeDescriptorContext? context, CultureInfo?
         // not parse the Python type to a CLR type, only to a new Python type.
         Type[] types = destinationType.GetGenericArguments();
         nint tupleSize = CPythonAPI.PyTuple_Size(pyObj);
-        object?[] clrValues = new object[Math.Max(8, tupleSize)];
+        object?[] clrValues = new object[Math.Min(8, tupleSize)];
```
tonybaloney, line 34: Can we `stackalloc` this? Just like we have in the `CallArgs` function

src/CSnakes.Runtime/PyObjectTypeConverter.Dictionary.cs
```diff
@@ -46,17 +45,15 @@ internal partial class PyObjectTypeConverter
...
-            int result = CPythonAPI.PyDict_SetItem(pyDict, ToPython(kvp.Key, context, culture), ToPython(kvp.Value, context, culture));
-            if (result == -1)
-            {
-                PyObject.ThrowPythonExceptionAsClrException();
-            }
-        }
-
-        return pyDict;
+        List<PyObject> keys = new();
+        List<PyObject> values = new();
```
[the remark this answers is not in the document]
tonybaloney: The Dispose callback won't work with a ref type like `Span<>` so it'll need to be an array. I'll do the first option for now.

tonybaloney:
@AaronRobinsonMSFT can you take a look at https://github.com/tonybaloney/CSnakes/pull/159/commits/89f712bbcfe29e82dc72337fb65afde2159c0872.
This uses TLS for the GIL state and makes it recursive like we discussed.

tonybaloney:
One of the major issues for Perf is the conversion from Python to .NET in collection types.
`ConvertTo` takes a `PyObject`.

For lists, tuples and dictionaries we unpack the items in the container and create a `PyObject` from the handle in each item then send it back through `ConvertTo` because it is recursive. 

In the profile app we're making 35 `PyObject` allocations _per_ call to the method. That number should be fewer than 10. 
Most of those allocations are done outbound, which is wrong. 

We could modify the signature for `ConvertTo` to use a PyObject handle (`IntPtr` or `nint`) as this is a value type. We'd need to track the releases and balance the references to avoid memory leaks. We currently handle this by creating a `PyObject` for every new ref with the guarantee that it will be disposed and decref'd

src/CSnakes.Runtime/CPython/Long.cs
```diff
@@ -16,6 +16,9 @@ internal unsafe partial class CPythonAPI
     [LibraryImport(PythonLibraryName)]
     internal static partial long PyLong_AsLongLong(PyObject p);
 
+    [LibraryImport(PythonLibraryName)]
+    internal static partial int PyLong_AsLong(PyObject p);
```
AaronRobinsonMSFT, line 20: Ah ha! This is part of a previous issue I had. This signature is only correct on Windows. Reading the [API](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong), this returns a C `long`. The details are sad and I'll just point you at the [documentation](https://learn.microsoft.com/dotnet/standard/native-interop/best-practices#cc-long).

src/CSnakes.Runtime/PyObjectTypeConverter.Tuple.cs
```diff
@@ -1,81 +1,75 @@
...
+        for (nint i = 0; i < tupleValues.Length; i++)
+        {
+            PyObject value = PyObject.Create(CPythonAPI.PyTuple_GetItemWithNewRef(pyObj, i));
+            tupleValues[i] = value;
+        }
+
+        // If we have more than 8 python values, we are going to have a nested tuple, which we have to unpack.
+        if (tupleValues.Length > 8)
+        {
+            // We are hitting nested tuples here, which will be treated in a different way.
```
AaronRobinsonMSFT, line 39: Unclear if this is a hot path, but this style of LINQ is a bit more expensive than a simple for loop and passing around a sliced `Span<object>`.

src/CSnakes.Runtime/PyObjectTypeConverter.cs
```diff
@@ -54,12 +51,12 @@ internal partial class PyObjectTypeConverter : TypeConverter
 
         if (destinationType == typeof(BigInteger) && CPythonAPI.IsPyLong(pyObject))
         {
-            return ConvertToBigInteger(pyObject, destinationType, context, culture);
+            return ConvertToBigInteger(pyObject, destinationType);
         }
 
         if (destinationType == typeof(int) && CPythonAPI.IsPyLong(pyObject))
```
AaronRobinsonMSFT, line 57: This predicate is going to have to get a bit more complicated.

src/CSnakes.Runtime/Python/GIL.cs
```diff
@@ -5,25 +5,50 @@ namespace CSnakes.Runtime.Python;
...
         public PyGilState()
         {
             Debug.Assert(CPythonAPI.IsInitialized);
             gilState = CPythonAPI.PyGILState_Ensure();
+            recursionCount = 1;
+        }
+
+        public void Ensure()
+        {
+            if (recursionCount == 0)
```
AaronRobinsonMSFT, line 24: Nevermind my previous comment on this code. I see how this is being used now.

src/CSnakes.Runtime/CPython/Long.cs
tonybaloney: I'll handle that in another issue I think. I'd thought it was optimistic to use `long`

tonybaloney:
The initial benchmark:

Using the release package from main - 5269ms.
[d7d48a1](https://github.com/tonybaloney/CSnakes/pull/159/commits/d7d48a163af36d8fb363cbd88bc5a2c5f3b6aacc) - 2674ms 

So we've halved the benchmark run time so far.

I've added a second test that just returns `PyObject` and that executes in 866ms, so the conversion back to .NET types is costing about 75% of the execution time. This benchmark is a tuple within a dictionary within a list, so it's going to particularly highlight the performance of container conversions.

tonybaloney:
Implementing #165 took another 5% off:

```default
Program: Information: Took 2465ms for 100000 cycles
Program: Information: Took 826ms for 100000 cycles without conversions
```

PR #178, opened 2024-08-29
Make the returned dictionary and list types lazy

tonybaloney:
This makes a major change to the Python Dictionary -> Managed Dictionary conversion introducing a lazy dictionary.

This Dictionary implements `IReadOnlyDictionary` but only yields the values when requested. It didn't make sense to eagerly convert the entire Python dictionary (recursively) as Python isn't doing that either.

This would be a performance improvement in almost all cases for dictionaries. 

As part of this change I also implemented an `Enumerator` class which uses the `iter()` protocol in Python to convert a Python list into a lazy enumerable in .NET. 

The Python List type now becomes an `IReadOnlyList` so that users can fetch a specific index.


tonybaloney:
Benchmark pass 1:

 Method                    | Mean        | Error     | StdDev    |
-------------------------- |------------:|----------:|----------:|
 ComplexReturn             |  4,991.8 ns |  40.60 ns |  37.98 ns |
 ComplexReturnLazy         |  3,756.9 ns |  52.38 ns |  43.74 ns |
 FunctionReturnsList       | 14,641.5 ns |  43.71 ns |  38.75 ns |
 FunctionTakesList         | 36,088.4 ns | 421.27 ns | 373.45 ns |
 FunctionTakesDictionary   | 68,186.7 ns | 904.15 ns | 845.74 ns |
 FunctionReturnsDictionary | 15,857.7 ns | 127.18 ns | 118.96 ns |
 FunctionReturnsTuple      |  1,207.8 ns |   4.40 ns |   3.90 ns |
 FunctionTakesTuple        |  1,992.0 ns |  38.37 ns |  35.89 ns |
 FunctionTakesValueTypes   |    801.5 ns |   1.41 ns |   1.10 ns |
 EmptyFunction             |    256.7 ns |   0.45 ns |   0.40 ns |

Improvements: 
Complex Return (2x improvement)
Function Returns Dictionary (~3x improvement)

src/Integration.Tests/IntegrationTestBase.cs
```diff
@@ -22,7 +22,8 @@ public IntegrationTestBase()
                 var pb = services.WithPython();
                 pb.WithHome(Path.Join(Environment.CurrentDirectory, "python"));
 
-                pb.FromNuGet(pythonVersionWindows)
```
tonybaloney, line 25: Dammit not again

tonybaloney:
Method                    | Mean        | Error     | StdDev    |
-------------------------- |------------:|----------:|----------:|
 ComplexReturn             |  4,128.3 ns |  48.70 ns |  43.17 ns |
 ComplexReturnLazy         |  3,609.7 ns |  31.79 ns |  28.18 ns |
 FunctionReturnsList       |  4,221.4 ns |  84.15 ns | 106.42 ns |
 FunctionTakesList         | 28,264.3 ns | 217.99 ns | 193.24 ns |
 FunctionTakesDictionary   | 68,396.5 ns | 866.16 ns | 810.20 ns |
 FunctionReturnsDictionary | 15,697.7 ns |  93.08 ns |  87.07 ns |
 FunctionReturnsTuple      |  1,199.3 ns |   2.06 ns |   1.72 ns |
 FunctionTakesTuple        |  1,965.5 ns |  38.59 ns |  45.94 ns |
 FunctionTakesValueTypes   |    803.5 ns |   2.81 ns |   2.63 ns |
 EmptyFunction             |    255.7 ns |   0.60 ns |   0.53 ns |

- Complex Return additional 20% improvement
- Function Returns List ~3-4x improvement

src/CSnakes.Runtime/PyObjectTypeConverter.List.cs
```diff
@@ -1,68 +1,22 @@
-using CSnakes.Runtime.CPython;
 using CSnakes.Runtime.Python;
 using System.Collections;
 using System.Runtime.InteropServices;
 
 namespace CSnakes.Runtime;
 internal partial class PyObjectTypeConverter
 {
-    private static ICollection ConvertToList(PyObject pyObject, Type destinationType)
+    private static object ConvertToList(PyObject pyObject, Type destinationType)
```
aaronpowell, line 8: Does this have to be `object`, can't we use `IReadOnlyList`?

aaronpowell (approved):

src/CSnakes.Runtime/PyObjectTypeConverter.List.cs
tonybaloney: There are only Generic versions of that interface and we don't know `T`, so no

PR #187, opened 2024-09-05
Convert ndarray scalars to Span<T> using the buffer protocol

tonybaloney:
This implements the Python Buffer protocol as a type and exposes the data for _most_ unmanaged types as either a scalar `Span<>` or 2D array `Span2D`. 

There is also an optional ReadOnlySpan interface.

This is a significant feature for numpy so that users dont need to convert ndarrays back into lists and then CSnakes marshals them.

It also works for `bytes` (as `ReadOnlySpan<byte>`) and `bytearray` as (`Span<byte>`).

It also opens up a bidirectional interface for exchanging low-level data via the span.


src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,60 @@
+﻿using CSnakes.Runtime.Python;
```
tonybaloney, line 1: @AaronRobinsonMSFT can you help me make this and `PyBuffer.cs` more sane?

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct Py_buffer
+    {
+        public IntPtr buf;
+        public IntPtr obj;
+        public nint len;
+        public Int32 @readonly;
+        public nint itemsize;
+        public char* format;
```
AaronRobinsonMSFT, line 24:
```suggestion
        public byte* format;
```

This is a native `char`, so 8 bits, not 16 bits.

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,60 @@
+﻿using CSnakes.Runtime.Python;
+using System.Runtime.InteropServices;
+
+namespace CSnakes.Runtime.CPython;
+internal unsafe partial class CPythonAPI
+{
+    private const int PyBUF_SIMPLE = 0;
```
AaronRobinsonMSFT, line 7:
```csharp
    [Flags]
    private enum PyBUF
    {
        Simple = 0,
        Writable = 0x1,
        Format = 0x4,
        ND = 0x8,
        Strides = 0x10 | ND,
        CContiguous = 0x20 | Strides,
        FContiguous = 0x40 | Strides,
        AnyContiguous = 0x80 | Strides,
    }
```

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,60 @@
...
+    private const int PyBUF_F_CONTIGUOUS = (0x0040 | PyBUF_STRIDES);
+    private const int PyBUF_ANY_CONTIGUOUS = (0x0080 | PyBUF_STRIDES);
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct Py_buffer
+    {
+        public IntPtr buf;
+        public IntPtr obj;
+        public nint len;
+        public Int32 @readonly;
```
AaronRobinsonMSFT, line 22: I get the why here, but it is generally better to use the keyword.

```suggestion
        public int @readonly;
```

src/CSnakes.Runtime/Python/IPyBuffer.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    Int64 Length { get; }
+
+    bool Scalar { get; }
+
+    Span<Int32> AsInt32Scalar();
+    Span<Int64> AsInt64Scalar();
+    Span<UInt32> AsUInt32Scalar();
+    Span<UInt64> AsUInt64Scalar();
+    public Span<float> AsFloatScalar();
```
AaronRobinsonMSFT, line 12:
```suggestion
    Span<int> AsInt32Scalar();
    Span<long> AsInt64Scalar();
    Span<uint> AsUInt32Scalar();
    Span<ulong> AsUInt64Scalar();
    public Span<float> AsFloatScalar();
```

Is there a reason `AsFloatScalar()` is marked `public`?

What about `AsDoubleScalar()`? Unsure if that is valid. I'm having trouble getting the valid values for the `format` field from the documentation.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public PyBuffer(PyObject exporter)
+    {
+        _buffer = CPythonAPI.GetBuffer(exporter);
+    }
+
+    public void Dispose()
+    {
+        CPythonAPI.ReleaseBuffer(_buffer);
+    }
```
AaronRobinsonMSFT, line 16: This isn't enough to support the `IDisposable` contract. is that what you want?

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public Int64 Length => _buffer.len;
+
+    public bool Scalar => _buffer.ndim == 0;
+
+    private unsafe void EnsureFormat(char format)
+    {
+        if (!_buffer.format[0].Equals(format))
```
AaronRobinsonMSFT, line 24:
```suggestion
    private unsafe void EnsureFormat(char format)
    {
        if (!_buffer.format[0] == (byte)format)
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            throw new InvalidOperationException("Buffer is not a scalar");
+        }
+    }
+
+    public unsafe Span<Int32> AsInt32Scalar()
+    {
+        EnsureScalar();
+        EnsureFormat('l'); // TODO: i is also valid
+        return new Span<Int32>((void*)_buffer.buf, (int)(Length / 4));
```
AaronRobinsonMSFT, line 42:
```suggestion
        return new Span<Int32>((void*)_buffer.buf, (int)(Length / sizeof(int)));
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public unsafe Span<UInt32> AsUInt32Scalar()
+    {
+        EnsureScalar();
+        EnsureFormat('L'); // TODO: I is also valid
+        return new Span<UInt32>((void*)_buffer.buf, (int)(Length / 4));
```
AaronRobinsonMSFT, line 49:
```suggestion
        return new Span<UInt32>((void*)_buffer.buf, (int)(Length / sizeof(uint)));
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public unsafe Span<Int64> AsInt64Scalar()
+    {
+        EnsureScalar();
+        EnsureFormat('q');
+        return new Span<Int64>((void*)_buffer.buf, (int)(Length / 8));
```
AaronRobinsonMSFT, line 56:
```suggestion
        return new Span<Int64>((void*)_buffer.buf, (int)(Length / sizeof(long)));
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public unsafe Span<UInt64> AsUInt64Scalar()
+    {
+        EnsureScalar();
+        EnsureFormat('Q');
+        return new Span<UInt64>((void*)_buffer.buf, (int)(Length / 8));
```
AaronRobinsonMSFT, line 63:
```suggestion
        return new Span<UInt64>((void*)_buffer.buf, (int)(Length / sizeof(ulong)));
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public unsafe Span<float> AsFloatScalar() {
+        EnsureScalar();
+        EnsureFormat('f');
+        return new Span<float>((void*)_buffer.buf, (int)(Length / 4));
```
AaronRobinsonMSFT, line 69:
```suggestion
        return new Span<float>((void*)_buffer.buf, (int)(Length / sizeof(float)));
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    private void EnsureScalar()
+    {
+        if (!Scalar)
+        {
+            throw new InvalidOperationException("Buffer is not a scalar");
+        }
+    }
+
+    public unsafe Span<Int32> AsInt32Scalar()
```
AaronRobinsonMSFT, line 38: Do you care about endianness of any of these operations? .NET always defers to the platform. I'm assuming CPython does the same.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    private void EnsureScalar()
+    {
+        if (!Scalar)
+        {
+            throw new InvalidOperationException("Buffer is not a scalar");
+        }
+    }
```
AaronRobinsonMSFT, line 36: Can this be done in the constructor?

src/CSnakes.Runtime/Python/PyBuffer.cs
AaronRobinsonMSFT: You also might want to reset `_buffer` to avoid duplicate releases?

```csharp
_buffer = default;
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,71 @@
+﻿using CSnakes.Runtime.CPython;
+
+namespace CSnakes.Runtime.Python;
+internal class PyBuffer : IPyBuffer
```
AaronRobinsonMSFT, line 4:
```suggestion
internal sealed class PyBuffer : IPyBuffer
```

src/CSnakes.Runtime/CPython/Buffer.cs
tonybaloney: Ah, I was trying to find a flags enum type! thanks

src/CSnakes.Runtime/Python/PyBuffer.cs
tonybaloney: By default, it uses the platform endianness. But, you can specify that as well, I'll try and work out how to test it.

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,64 @@
...
+    internal static void ReleaseBuffer(Py_buffer view)
+    {
+        PyBuffer_Release(&view);
+    }
+
+    [LibraryImport(PythonLibraryName)]
+    private static partial int PyObject_CheckBuffer(PyObject ob);
+
+    [LibraryImport(PythonLibraryName)]
+    private static partial int PyObject_GetBuffer(PyObject ob, Py_buffer* view, int flags);
```
AaronRobinsonMSFT, line 60:
```suggestion
    private static partial int PyObject_GetBuffer(PyObject ob, Py_buffer* view, PyBUF flags);
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,113 @@
...
+        if (_format.Length == 0)
+        {
+            return ByteOrder.Native;
+        }
+        return Enum.TryParse(_format[0].ToString(), out ByteOrder byteOrder) ? byteOrder : ByteOrder.Native;
+    }
+
+    private void EnsureFormat(char format)
+    {
+        if (!_format.Contains(format))
```
AaronRobinsonMSFT, line 79: The `StartsWith()` function is already heavily optimized. It might be faster that `Contains()`.

AaronRobinsonMSFT: You could also do `_format[0] == format`?

src/CSnakes.Runtime/Python/IPyBuffer.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace CSnakes.Runtime.Python;
+public interface IPyBuffer
+{
+    Int64 Length { get; }
+
+    bool Scalar { get; }
+
+    Span<int> AsInt32Scalar();
```
AaronRobinsonMSFT, line 8: As a curiosity, are `byte`/`sbyte` or `short`/`ushort` interesting?

src/CSnakes.Runtime/Python/PyBuffer.cs
AaronRobinsonMSFT: Oh, I see. The format can have a lot more details.

src/CSnakes.Runtime/Python/IPyBuffer.cs
tonybaloney: Yes

tonybaloney:
Linux tests are failing because of #189 

Rest of the implementation I'm happy with

src/CSnakes.Runtime/Python/PyBuffer.cs
tonybaloney: I've just put a guard in there for now

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,64 @@
...
+        public nint* shape;
+        public nint* strides;
+        public nint* suboffsets;
+        public nint* @internal;
+    }
+
+    public static bool IsBuffer(PyObject p)
+    {
+        return PyObject_CheckBuffer(p) == 1;
+    }
```
aaronpowell, line 39:
```suggestion
    public static bool IsBuffer(PyObject p) => PyObject_CheckBuffer(p) == 1;
```

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            throw PyObject.ThrowPythonExceptionAsClrException();
+        }
+        return view;
+    }
+
+    internal static void ReleaseBuffer(Py_buffer view)
+    {
+        PyBuffer_Release(&view);
+    }
```
aaronpowell, line 54:
```suggestion
    internal static void ReleaseBuffer(Py_buffer view) => PyBuffer_Release(&view);
```

src/CSnakes.Runtime/Python/IPyBuffer.cs
aaronpowell, line 1: I kind of feel like the default should be a `ReadOnlySpan` and we should have an alternative like `AsBidirectionSpan`, so `AsByeSpan` would return `ReadOnlySpan<byte>` and `AsBidirectionalByteSpan` returns `Span<byte>`.

This means that by-default you can't futz with the memory that Python provides and then makes it a lot clearer that if you want to have something that is bidirectional you can have it.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,286 @@
...
+        {
+            throw new InvalidOperationException($"Buffer is not a {format}, it is {_format}");
+        }
+    }
+
+    private void EnsureFormat(Format format)
+    {
+        EnsureFormat((char)format);
+    }
```
aaronpowell, line 102:
```suggestion
    private void EnsureFormat(Format format) => EnsureFormat((char)format);
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -0,0 +1,286 @@
...
+    {
+        if (Dimensions != dimensions)
+        {
+            throw new InvalidOperationException($"Buffer is not {dimensions}D");
+        }
+    }
+
+    private unsafe void EnsureShapeAndStrides()
+    {
+        if (_buffer.shape == null || _buffer.strides == null)
```
aaronpowell, line 122:
```suggestion
        if (_buffer.shape is null || _buffer.strides is null)
```

aaronpowell (changes requested):
As I look at the implementation of `AsSpan` (and friends) it makes me rethink the overall approach to how we request the values back from Python. We rely a lot on throwing exceptions (and this more so than others), and that makes me wonder if we should move away from explicit return types and instead us a `TryGetXXX` pattern.

So the `AsSpan` would become `bool TryGetAsSpan<T>(out T span)`. Doing this allows for easier defensive coding, you request the value _conditionally_ and handle the failure to resolve it. The thing that you would lose is the _reason_ for failure, since you don't get an exception type which you can handle.

AaronRobinsonMSFT:
>  instead us a `TryGetXXX` pattern.
> Doing this allows for easier defensive coding, you request the value _conditionally_ and handle the failure to resolve it.

I'm not sure I agree. Applying the "Try" pattern is when there is a genuine chance of failure. A lot of times it comes down to parsing strings. In that case it is not uncommon that errors creep in. In this case the scenario is interop and data coming through is going to be handled during development - once it is "correct" it isn't going to be wrong. In that case, the current approach is far better as it is far less common that one needs to be defensive in production.

aaronpowell:
Hmm I guess you're right there. Maybe it'll be better to create a custom exception type here that has an enum representing the failure reason, rather than trying to parse the exception message to determine the course of action to take.

aaronpowell (approved):

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,58 @@
...
+        CContiguous = 0x20 | Strides,
+        FContiguous = 0x40 | Strides,
+        AnyContiguous = 0x80 | Strides,
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct Py_buffer
+    {
+        public IntPtr buf;
+        public IntPtr obj;
```
RussKie, line 24: any reason for these not be a `nint`?
```suggestion
        public nint buf;
        public nint obj;
```

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,58 @@
...
+        public nint len;
+        public nint itemsize;
+        public int @readonly;
+        public Int32 ndim;
```
RussKie, line 28:
```suggestion
        public int ndim;
```

src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -0,0 +1,58 @@
+﻿using CSnakes.Runtime.Python;
+using System.Runtime.InteropServices;
+
+namespace CSnakes.Runtime.CPython;
+internal unsafe partial class CPythonAPI
+{
+    [Flags]
+    private enum PyBUF
```
RussKie, line 8:
```suggestion
    private enum PyBUF : int
```

docs/buffers.md
```diff
@@ -0,0 +1,104 @@
...
+
+import numpy as np
+
+def example_array() -> Buffer:
+    return np.array([True, False, True, False, False], dtype=np.bool_)
+```
+
+In this example, the `example_array` function returns a NumPy array of boolean values. The `Buffer` type hint indicates that the function returns an object that supports the Buffer Protocol.
+
+From C#, CSnakes will return an implementation of `CSnakes.Runtime.Python.IPyBuffer` that wraps the buffer and provides access to the data as a Span.
```
aaronpowell, line 26:
```suggestion
From C#, CSnakes will return an `CSnakes.Runtime.Python.IPyBuffer` that provides access to the data as a `Span`.
```

docs/buffers.md
```diff
@@ -0,0 +1,104 @@
...
+if (bufferObject.IsScalar) {
+
+// Get the buffer contents as a Span of bool
+	Span<bool> result = bufferObject.AsBoolSpan();
+	Console.WriteLine(result[0]); // True
+	Console.WriteLine(result[4]); // False
+}
+```
+
+Since Span is writeable, you can also modify the buffer contents from C# and the changes will be reflected in the Python object.
```
aaronpowell, line 44: I feel like this should be more called out in big flashing red lights. The fact that it's writable, allowing you to write to memory you don't "own" is a very dangerous thing.

Maybe we should call out that it's recommended that you use the `ReadOnly` version unless you have a very explicit need.

docs/buffers.md
```diff
@@ -0,0 +1,104 @@
...
+// Where testModule is a Python module that contains the example_array_2d function
+var result = testModule.ExampleArray2D();
+
+// Get the buffer contents as a two-dimensional Span of int
+Span2D<int> result2D = result.AsIntSpan2D();
+Console.WriteLine(result2D[0, 0]); // 1
+
+```
+
+var bufferObject = testModule.ExampleArray2D();
```
aaronpowell, line 77: Mistakenly left there?

docs/buffers.md
```diff
@@ -0,0 +1,104 @@
...
+
+Since Span is writeable, you can also modify the buffer contents from C# and the changes will be reflected in the Python object.
+
+If you want a read-only view of the buffer, you can use the `As[T]ReadOnly` method to get a read-only Span.
+
+```csharp
+Span<bool> result = bufferObject.AsBoolReadOnlySpan();
+```
+
+## Two-Dimensional Buffers
```
aaronpowell, line 52: Again, we should call out that buffers are writable

aaronpowell, line 52: How do we do mixed-type 2D buffers, or is that not possible in Python?

I'm thinking something like: `[[1, 1.1], [2, 2.2]]`, where the types are `int` and `float`.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -70,11 +70,58 @@ public void Dispose()
...
+    public Type GetItemType()
+    {
+        // The format string contains the type of the buffer, normally in the first
+        // position, but the first character can also be the byte order.
+        for (int i = 0; i < _format.Length; i++)
+        {
+            if (Enum.IsDefined(typeof(Format), (int)_format[i]))
+            {
+                var format = (Format)_format[i];
+                switch (format)
```
aaronpowell, line 88: We can reduce this down with a switch expression

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -85,7 +132,15 @@ private ByteOrder GetByteOrder()
...
-        return Enum.TryParse(_format[0].ToString(), out ByteOrder byteOrder) ? byteOrder : ByteOrder.Native;
+
+        if (Enum.IsDefined(typeof(ByteOrder), (int)_format[0]))
+        {
+            return (ByteOrder)_format[0];
+        }
+        else
+        {
+            return ByteOrder.Native;
+        }
```
aaronpowell, line 143:
```suggestion
        return ByteOrder.Native;
```

src/Integration.Tests/Integration.Tests.csproj
```diff
@@ -3,67 +3,8 @@
   <PropertyGroup>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
-	<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
-</PropertyGroup>
-
-  <ItemGroup>
```
aaronpowell: This isn't needed anymore?

aaronpowell (changes requested):

docs/buffers.md
tonybaloney: That's not possible in numpy. NDArrays are packed typed arrays with C types inside them

aaronpowell (approved):

PR #201, opened 2024-09-17
Parser supports additional syntax for type arguments

### LINKED ISSUE #170: Callable[[args], ret] breaks the parser
Related to #169 

https://peps.python.org/pep-0677/

The following syntax breaks the parser:

```python
def f(a: Callable[[str], int]) -> None:
  ...
```

tonybaloney:
Fixes #170


aaronpowell (approved):

src/CSnakes/Parser/PythonParser.TypeDef.cs
```diff
@@ -14,19 +14,21 @@ public static partial class PythonParser
...
+                    Token.EqualTo(PythonToken.Comma),
+                    Token.EqualTo(PythonToken.CloseBracket)
+                )
+        )
 #pragma warning restore CS8620
-            .OptionalOrDefault()
-         select new PythonTypeSpec(name.ToStringValue(), openBracket))
-        .Named("Type Definition");
+        .OptionalOrDefault()
+     select name.HasValue ? new PythonTypeSpec(name.ToStringValue(), openBracket)
```
tonybaloney, line 30: @aaronpowell  worked out a clean solution.

PR #194, opened 2024-09-16
Use CI version of Python

RussKie:
Resolves #154

PR #202, opened 2024-09-17
Support .NET 9

RussKie:
## Changes

* Revert #194 "Use CI version of Python"
To allow samples to be self-contained, duplicate the code

* Convert to CPM and add support for net9.0 TFM

* Refactor packaging
    - Move packaging related definitions into props and targets.
    - Rework how CSnakes nupkg is created.
    - Align naming of the locally generated packages with the CI-produced packages - i.e., local packages will now have `-dev.123.123` suffix (e.g., `CSnakes.1.0.17-dev.240917.1659.nupkg`). This streamlines the local devex for the samples projects.

## Results

image

image


aaronpowell (commented):
I didn't realise you could ship multiple TFM targets in a single NuGet package, nice!

I also had wanted to move to CPM, thanks for tackling that.

With CI - do we need to drop the version suffix that we add from the GitHub Action now that it's generating it in the `Package.targets`?

RussKie:
Right now, there's only one small issue - the CI tests only report status for one TFM, though both TFMs are tested. I confirm this by inspecting the published binlogs. 
I don't think is a blocking issue, and it can be corrected later.
I'm chatting with the Test folks to find out how this can be corrected.

RussKie:
> With CI - do we need to drop the version suffix that we add from the GitHub Action now that it's generating it in the `Package.targets`?

No. My change is aimed at improving the local devex, and it names locally-built nupkgs in the way the Samples expect - `CSnakes.Runtime.1.0.17-dev.240918-1054.nupkg`. 
This branding only occurs locally (i.e., not on CI).

tonybaloney (approved):

PR #204, opened 2024-09-17
Reuse `PyEnumerable`, removing need for `PyKeyValuePairEnumerable`

atifaziz:
`PyKeyValuePairEnumerable<,>` had almost an identical implementation to `PyEnumerable<>` except for the conversion of the `PyObject` value. The former calls `PyObject.As<,>`:

https://github.com/tonybaloney/CSnakes/blob/b9a6b8d76c5de5c3beb2b46d1639b06253e4002d/src/CSnakes.Runtime/Python/PyKeyValuePairEnumerable.cs#L42-L43

whereas the latter calls `PyObject.As<>`:

https://github.com/tonybaloney/CSnakes/blob/b9a6b8d76c5de5c3beb2b46d1639b06253e4002d/src/CSnakes.Runtime/Python/PyEnumerable.cs#L42-L43

That's a lot of duplication of maintain, sync, test and cover. This PR consolidates the duplication so there's only one bulk implementation. It does this by adding another type parameter (called `TImporter`) to `PyEnumerable<>` that determines how the `PyObject` is imported (into presumably a managed type):

```c#
internal class PyEnumerable<TValue, TImporter> : IEnumerable<TValue>, IEnumerator<TValue>, IDisposable
    where TImporter : IPyObjectImporter<TValue>
{
    // omitted for brevity
}
```

The importer has a _static_ member, which bears _zero cost_ (that is, no allocation and no virtual dispatch needed):

```c#
interface IPyObjectImporter<out T>
{
    static abstract T Import(PyObject pyObj);
}
```

The two import strategies, `PyObject.As<>` and `PyObject.As<,>`, are then embodied into implementations of this interface. The import strategy is chosen at the time `PyEnumerable<,>` is instantiated. A `PyEnumerable<>` with a single type parameter still exists, but which uses the default strategy of `PyObject.As<>` so most of the code is untouched except in `PyDictionary<,>.GetEnumerator`.


tonybaloney (approved):

tonybaloney:
Love it! this is a lot cleaner

PR #207, opened 2024-09-18
Adding F# sample

aaronpowell:


tonybaloney:
Does this get built by CI?

aaronpowell:
Yep, it's in the same solution as the other "basic" samples

PR #210, opened 2024-09-18
Supporting forks in CI better

### LINKED ISSUE #208: CI builds fails for forks
I work off a forked codebase. All my PRs fail on the following runs:

* CI / test-reporting:
image

* publish-github-packages
image

aaronpowell:
Some of our CI jobs require access to the `GITHUB_TOKEN`, but the problem is that when they are run from a fork, they don't have permission to do what we need done.

To address this, we need to split the jobs that need permissions out into a new workflow which is run once the triggered workflow completes, since that trigger will run in context of the repo and mean that we can use the `GITHUB_TOKEN`.

Fixes #208


RussKie (approved):
:shipit:

PR #211, opened 2024-09-18
Rename CSnake project to CSnake.SourceGeneration and unify nupkgs

### LINKED ISSUE #209: Rename CSnake project to CSnake.SourceGeneration
It is customary for Roslyn source generator and analyzer projects to have "Generator" and "Analyzers" suffixes (or variations of thereof).

image

To align we should rename CSnake project to CSnake.SourceGeneration

### LINKED ISSUE #290: Load Python from String or File Path
I apologize if I've overlooked this. I am trying out this project to compare to other options as the support and stated goals seem outstanding (saw on .NET video on YouTube). Congrats on the project and visibility!

What I'd like to be able to do for our project is load Python code from a string or an arbitrary file that is not present at build time. This is for a general scripting engine where we allow JavaScript and Python.

While the code generation is awesome and we may use it for future solutions, that's not what we are aiming to use at this time. I see from the docs that you can in fact bypass codegen, but it still seems to load a module by name rather than by file path or passing in a string of Python.

Is this currently possible or on the roadmap? Or is that outside the scope of this project. Thanks for the project!

RussKie:
* Rename CSnake project to CSnake.SourceGeneration (Resolves #209)
    image

* Repackage the source generator into the runtime package
    image
    The SG can be disabled by the consumer project by settings the following MSBuild property:
    ```xml
    <DisableCSnakesRuntimeSourceGenerator>true</DisableCSnakesRuntimeSourceGenerator>
    ```


RussKie:
The Samples CI fails because there's no CSnake.SourceGeneration package exists on the nuget feed.

tonybaloney (approved):

tonybaloney:
Looks like this also reformats a lot of the source code. it will introduce a lot of merge conflicts. Suggest we check what else is in the backlog first

RussKie:
> Looks like this also reformats a lot of the source code. it will introduce a lot of merge conflicts. Suggest we check what else is in the backlog first

I've moved files due to the folder rename. Git should handle this just fine.

tonybaloney:
Looks like there are some merge conflicts

RussKie:
> Looks like there are some merge conflicts

Resolved

RussKie:
The Samples are failing because they require the new nupkg with the SG packaged in.

src/CSnakes.SourceGeneration/CaseHelper.cs
```diff
@@ -1,16 +1,16 @@
...
+            return string.Join("", snakeCase.Split('_').Select(s => s.Length > 1 ? char.ToUpperInvariant(s[0]) + s.Substring(1) : "_"));
+        }
+
+        public static string ToLowerPascalCase(this string snakeCase)
+        {
+            // Make sure the first letter is lowercase
+            return char.ToLowerInvariant(snakeCase[0]) + ToPascalCase(snakeCase).Substring(1);
+        }
+    }
+}
```
aaronpowell, line 16:
```suggestion
﻿namespace CSnakes;
public static class CaseHelper
{
    public static string ToPascalCase(this string snakeCase)
    {
        return string.Join("", snakeCase.Split('_').Select(s => s.Length > 1 ? char.ToUpperInvariant(s[0]) + s.Substring(1) : "_"));
    }

    public static string ToLowerPascalCase(this string snakeCase)
    {
        // Make sure the first letter is lowercase
        return char.ToLowerInvariant(snakeCase[0]) + ToPascalCase(snakeCase).Substring(1);
    }
}
```

aaronpowell (commented):
Is it a single NuGet package now called `CSnakes.Runtime`?

src/CSnakes.SourceGeneration/CaseHelper.cs
aaronpowell: Using this as an opportunity to fix the namespace to be file-scoped

RussKie: This would be a diff-breaking change.
I suggest performing this change as a separate PR **with no other big PR in flight**.

RussKie:
> Is it a single NuGet package now called `CSnakes.Runtime`?

Correct. See the screenshot above

aaronpowell (approved):

PR #220, opened 2024-09-19
Fix LD_PRELOAD workarounds

### LINKED ISSUE #112: Fix virtual environments on Linux
https://github.com/tonybaloney/CSnakes/blob/0079b2baa8a1212b0b85559c322d611e45c07e8a/src/Integration.Tests/TestDependency.cs#L9

```csharp

﻿using Integration.Tests;
using System.Runtime.InteropServices;

public class TestDependency : IntegrationTestBase
{
    [Fact]
    public void VerifyInstalledPackage()
    {
        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) // TODO: Fix virtual environments on Linux
        {
            Assert.True(Env.TestDependency().TestNothing());
        }
    }
}

```

tonybaloney:
.Net LoadLibrary doesn't offer a way of setting the RTLD flags for Linux.

Python needs global symbols so that further dynamically loaded libraries can resolve symbols from the Python API. 

Our current workaround is to LD_PRELOAD the `libpython` so in the devcontainer and the tests. But this isn't practical for all users.

Fixes #112

### DIFF
--- a/src/CSnakes.Runtime/CPython/Init.cs
+++ b/src/CSnakes.Runtime/CPython/Init.cs
@@ -35,11 +35,11 @@ internal unsafe partial class CPythonAPI : IDisposable
     private enum RTLD : int
     {
         LOCAL = 0,
-        LAZY  = 1,
+        LAZY = 1,
         NOW = 2,
-        NOLOAD=4,
-        DEEPBIND=8,
-        GLOBAL=0x00100
+        NOLOAD = 4,
+        DEEPBIND = 8,
+        GLOBAL = 0x00100
     }
 
 
--- a/src/CSnakes.Runtime/CPython/Init.cs
+++ b/src/CSnakes.Runtime/CPython/Init.cs
@@ -45,7 +45,6 @@ internal unsafe partial class CPythonAPI : IDisposable
     [LibraryImport("libdl.so.2", StringMarshalling = StringMarshalling.Utf8)]
     private static partial nint dlopen(string path, int flags);
 
-
     private static IntPtr DllImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
     {
         if (libraryName == PythonLibraryName)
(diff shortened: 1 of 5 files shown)


aaronpowell (approved):

src/CSnakes.Runtime/CPython/Init.cs
```diff
@@ -31,10 +31,30 @@ public CPythonAPI(string pythonLibraryPath, Version version)
...
+
+    [LibraryImport("libdl.so.2", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial nint dlopen(string path, int flags);
+
+
     private static IntPtr DllImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
     {
         if (libraryName == PythonLibraryName)
         {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)){
```
AaronRobinsonMSFT, line 54: I would add an explicit comment on why Linux is a special case here.

src/CSnakes.Runtime/CPython/Init.cs
```diff
@@ -31,10 +31,30 @@ public CPythonAPI(string pythonLibraryPath, Version version)
...
+    private enum RTLD : int
+    {
+        LOCAL = 0,
+        LAZY = 1,
+        NOW = 2,
+        NOLOAD = 4,
+        DEEPBIND = 8,
+        GLOBAL = 0x00100
+    }
+
```
AaronRobinsonMSFT, line 44:
```suggestion

```

src/CSnakes.Runtime/CPython/Init.cs
```diff
@@ -31,10 +31,30 @@ public CPythonAPI(string pythonLibraryPath, Version version)
...
+
+    [LibraryImport("libdl.so.2", StringMarshalling = StringMarshalling.Utf8)]
+    private static partial nint dlopen(string path, int flags);
+
```
AaronRobinsonMSFT, line 48:
```suggestion

```

AaronRobinsonMSFT (approved):

PR #226, opened 2024-09-25
Recursive search of the folder "ExamplePythonDependency" in Program.cs, not working in CodeSpaces

elbruno:


aaronpowell:
Do you know why it wasn't working in a code space? It should run the same from path discovery as other OS environments.

samples/simple/QuickConsoleTest/Program.cs
```diff
@@ -9,8 +9,9 @@
 var builder = Host.CreateDefaultBuilder(args)
     .ConfigureServices(services =>
     {
-        var home = Path.Join(Environment.CurrentDirectory, "..", "..", "..", "..", "ExamplePythonDependency");
```
elbruno, line 12: This path resolve line is not working while running in a GitHub Codespace.
Best

Issue #183, opened 2024-08-31
Adding env support for conda


crazy4pi314:
I have gotten conda envs to work with csnakes by specifying hard paths with the `FromFolder` and `WithVirtualEnvironment` options on the source builder. I wanted to document the different ways I think these paths can be found programmatically from the [conda docs](https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html) with example output:

```
> echo $env:CONDA_PREFIX
C:\Users\sarahkaiser\Anaconda3

> conda env list --json
{
  "envs": [
    "C:\\Users\\sarahkaiser\\Anaconda3",
    "C:\\Users\\sarahkaiser\\Anaconda3\\envs\\csnakes",
    "C:\\Users\\sarahkaiser\\Anaconda3\\envs\\distats-bot",
    "C:\\Users\\sarahkaiser\\micromambaenv\\envs\\dask-tutorial-advanced",
    "c:\\Users\\sarahkaiser\\Anaconda3\\envs\\ghstats"
  ]
}
```
You will note that conda does list envs that are at a different base prefix since I also have micromamba installed.

If the env I was looking for was csnakes it would be at:
```
> $(conda info --base)/envs/csnakes
```

tonybaloney:
Implemented in #232

PR #232, opened 2024-10-06
Add Conda Environment Management and Locator Support

tonybaloney:
- abstracts environment management (as we might want to add `uv` in future as well)
- moves venv logic into a standalone class
- refactors python builder to locate environment management class
- implements a conda locator for the python provided by conda (effectively a folder locator for now)
- implements a `WithCondaEnvironment` for adding a named environment
- implements a conda installer for installing conda packages from a `environment.yml` file

Implements #183

### DIFF
--- a/.github/workflows/dotnet-ci.yml
+++ b/.github/workflows/dotnet-ci.yml
@@ -20,7 +20,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-
+      - uses: conda-incubator/setup-miniconda@v3
+        id: setup-conda
+        with:
+          auto-update-conda: true
+          python-version: ${{ matrix.python-version }}
       - name: Setup Python
         id: installpython
         uses: actions/setup-python@v5
--- a/.github/workflows/dotnet-ci.yml
+++ b/.github/workflows/dotnet-ci.yml
@@ -25,6 +25,8 @@ jobs:
         with:
           auto-update-conda: true
           python-version: ${{ matrix.python-version }}
+      - name: cleanup conda-incubator/setup-miniconda
+        run: conda clean --all --yes
       - name: Setup Python
         id: installpython
         uses: actions/setup-python@v5
--- a/src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
+++ b/src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
@@ -1,22 +1,9 @@
 ﻿using CSnakes.Runtime.Locators;
 using Microsoft.Extensions.Logging;
-using System.Diagnostics;
-
 
 namespace CSnakes.Runtime.EnvironmentManagement;
-internal class CondaEnvironmentManagement : IEnvironmentManagement
+internal class CondaEnvironmentManagement(string name, bool ensureExists, CondaLocator conda, string environmentSpecPath) : IEnvironmentManagement
 {
-    private readonly string name;
-    private readonly bool ensureExists;
-    private readonly CondaLocator conda;
-
-    internal CondaEnvironmentManagement(string name, bool ensureExists, CondaLocator conda)
-    {
-        this.name = name;
-        this.conda = conda;
-        this.ensureExists = ensureExists;
-    }
-
     public void EnsureEnvironment(ILogger logger, PythonLocationMetadata pythonLocation)
     {
         if (!ensureExists)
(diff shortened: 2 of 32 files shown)


src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
```diff
@@ -0,0 +1,37 @@
+﻿using CSnakes.Runtime.Locators;
+using Microsoft.Extensions.Logging;
+
+namespace CSnakes.Runtime.EnvironmentManagement;
+internal class CondaEnvironmentManagement(string name, bool ensureExists, CondaLocator conda, string? environmentSpecPath) : IEnvironmentManagement
```
aaronpowell, line 5: Is there a need for `environmentSpecPath` if it's not used?

src/CSnakes.Runtime/EnvironmentManagement/VenvEnvironmentManagement.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Diagnostics;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace CSnakes.Runtime.EnvironmentManagement;
+internal class VenvEnvironmentManagement : IEnvironmentManagement
+{
+    private readonly string path;
+    private readonly bool ensureExists;
+
+    internal VenvEnvironmentManagement(string path, bool ensureExists)
```
aaronpowell, line 13: Let's be consistent and make this a primary constructor like other types.

src/CSnakes.Runtime/EnvironmentManagement/VenvEnvironmentManagement.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        this.path = path;
+        this.ensureExists = ensureExists;
+    }
+
+    public void EnsureEnvironment(ILogger logger, PythonLocationMetadata pythonLocation)
```
aaronpowell, line 19: Should `ILogger` be provided to the constructor?

src/CSnakes.Runtime/EnvironmentManagement/VenvEnvironmentManagement.cs
```diff
@@ -0,0 +1,56 @@
+﻿using CSnakes.Runtime.Locators;
+using Microsoft.Extensions.Logging;
+using System.ComponentModel;
+using System.Diagnostics;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
aaronpowell, line 5: What are we pulling from here?

src/CSnakes.Runtime/Locators/CondaLocator.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    internal (Process process, string? output, string? errors) ExecuteCondaCommand(string arguments)
+    {
+        return ProcessUtils.ExecuteCommand(logger, condaBinaryPath, arguments);
+    }
+
+    internal bool ExecuteCondaShellCommand(string arguments)
+    {
+        return ProcessUtils.ExecuteShellCommand(logger, condaBinaryPath, arguments);
+    }
```
aaronpowell, line 54:
```suggestion
    internal (Process process, string? output, string? errors) ExecuteCondaCommand(string arguments) => ProcessUtils.ExecuteCommand(logger, condaBinaryPath, arguments);

    internal bool ExecuteCondaShellCommand(string arguments) => return ProcessUtils.ExecuteShellCommand(logger, 
    }
```

src/CSnakes.Runtime/ProcessUtils.cs
```diff
@@ -0,0 +1,80 @@
+﻿using CSnakes.Runtime.Locators;
+using Microsoft.Extensions.Logging;
+using System.Diagnostics;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace CSnakes.Runtime
```
aaronpowell, line 6: Use a file-scoped namespace.

aaronpowell (changes requested):

src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
tonybaloney: yes, because it'll be used in future. `ensureExists` isn't implemented but it'll need that parameter and I don't want to change the API when it is implemented. For now you get a not supported error

.github/workflows/dotnet-ci.yml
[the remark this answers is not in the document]
tonybaloney: this is the only workflow that runs the tests, which is where it's required.

aaronpowell (approved):

PR #234, opened 2024-10-06
Refactor `PythonConstant` types into a true union

atifaziz:
This PR proposes to refactor `PythonConstant` types into a true union via a class hierarchy where `PythonConstant` becomes an abstract base class and the various types its nested subclasses. This has the following benefits:

- Each subclass has its own clear, simpler and overriding implementation of `ToString` instead of a switch-case previously.
- Each subclass holds its own strong-typed value, e.g. `Value` on `PythonConstant.Integer` is typed as `long` whereas `Value` on `PythonConstant.String` is typed as `string`.
- Each subclass only requires the storage space for its value instead of the sum of all possible values.
- Subclasses like `Bool` and `None` have statically initialised values, which avoids unnecessary allocations.
- `PythonConstant.ConstantType` is no longer needed.
- Subclasses enable simpler, stronger and more robust pattern-matching (see updates in `ArgumentReflection.ArgumentSyntax`).

### DIFF
--- a/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
+++ b/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
@@ -44,59 +44,59 @@ public static partial class PythonParser
         from close in Character.EqualTo('\'')
         select Unit.Value;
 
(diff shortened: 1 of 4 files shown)


tonybaloney (approved):

tonybaloney:
Thanks, this is a lot cleaner. I think we started with 2 constant types then it got out of hand quickly

PR #235, opened 2024-10-06
Use a union struct for an optimal `PythonConstant`

atifaziz:
This PR is an alternative design to the refactoring proposed in PR #234. Only one of the two should be merged.

In contrast to PR #234, which uses a type hierarchy of subclasses to represent a union of all `PythonConstant` types, this uses a single `struct` as a union. It has the following benefits with respect to the current implementation and the one proposed in PR #234:

- It uses a single storage for all scalar value types (`bool`, `long` and `double`) and another for strings, instead of the sum of storage needed for all value types.
- Being a value type, it requires no additional heap allocations itself.
- In its default state (`default(PythonConstant)`), it represents `None`.
- Various properties provide strong-typed access to the underlying value (union of scalars or string) and designed to be friendly for pattern-matching (see `ArgumentReflection.ArgumentSyntax` for application). In fact, `PythonConstant.ConstantType` can even be made private since it's no longer used outside of `PythonConstant`.

The main motivation for accepting this design over PR #234 is zero heap footprint.

### DIFF
--- a/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
+++ b/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
@@ -68,13 +68,13 @@ public static partial class PythonParser
 
     public static TokenListParser<PythonToken, PythonConstant> HexidecimalIntegerConstantTokenizer { get; } =
         Token.EqualTo(PythonToken.HexidecimalInteger)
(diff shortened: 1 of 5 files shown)


tonybaloney:
> The main motivation for accepting this design over PR https://github.com/tonybaloney/CSnakes/pull/234 is zero heap footprint.

thanks. The memory consumption of the parser for the source generation is low-priority. 

I'm having a hard time understanding the code on this one and prefer the other PR. Granted, my C# knowledge is intermediate (a way from advanced).

Issue #257, opened 2024-10-11
Add API for reloading modules


tonybaloney:
This will be useful in future, but in the generated code, we import the module (and can optionally dispose it). But Python has an API for reloading modules.

This would provide an easy way to hot-reloading without restarting the .NET process.

PR #259, opened 2024-10-11
Add reload module functionality and connect generated classes to Hot Reload

tonybaloney:
Implements #257


tonybaloney:
FYI @atifaziz

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -119,8 +141,13 @@ public void Dispose()
             }
             public interface I{{pascalFileName}} : IDisposable
             {
+                /// <summary>
+                /// Reload the module, useful for development if the Python code has changed.
+                /// </summary>
+                public void ReloadModule();
```
atifaziz, line 147: This ought to be inherited through an interface in `CSnakes.Runtime` along the lines of:

```c#
public interface IReloadable : IDisposable
{
    /// <summary>
    /// Reload the module, useful for development if the Python code has changed.
    /// </summary>
    public void ReloadModule();
}
```

rather than being generated. The name is up for debate. While the chances are super low, this will also help to avoid the potential for a compilation error if the source Python module also contains a top-level function named `reload_module`; the implementing can [explicitly implement](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation) the interface:

```c#
void IModuleProxy.ReloadModule() { /* ... */ }
```

In the absolute worst case, the user will get the [CS0108](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0108) warning instead of a compilation error. In fact, the source generator could add the `new` keyword if such a conflict exists to avoid a CS0108 warning altogether.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -107,6 +117,18 @@ private class {{pascalFileName}}Internal : I{{pascalFileName}}
...
 
+                    /// <summary>
+                    /// Reload the module, useful for development if the Python code has changed.
+                    /// </summary>
+                    public void ReloadModule() {
+                        logger.LogDebug("Reloading module {ModuleName}", "{{fileName}}");
+                        using (GIL.Acquire())
+                        {
+                            module = Import.ReloadModule(module);
+                            {{string.Join(Environment.NewLine, functionNames.Select(f => $"this.{f.Field} = module.GetAttr(\"{f.Attr}\");"))}}
```
atifaziz, line 128: This could go in a separate method that's shared with the same code being generated in the constructor so it's JIT'd once.

More importantly though, I see a few concerns here:

- Won't this leak the previous function objects since the fields just get overwritten with potentially new objects?
- Is it exception-safe? What if the initialisation of the fields fails halfway? Won't it leave `this` in some strange state?
- What happens if a function was removed, updated (the signature rather than the code) or added?

I think the supported scenarios should be carefully documented to keep expectations in check.

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+public class ImportTests : RuntimeTestBase
+{
+    [Fact]
+    public void TestImportModule()
+    {
+        using (GIL.Acquire())
+        {
+            using PyObject? sys = Import.ImportModule("sys");
+            Assert.NotNull(sys);
+            Assert.Equal("<module 'sys' (built-in)>", sys!.ToString());
```
atifaziz, line 14: The `!` isn't needed since [`Assert.NotNull` narrows its argument to being non-nullable](https://github.com/xunit/assert.xunit/blob/cac8b688c193c0f244a0bedf3bb60feeb32d377a/NullAsserts.cs#L31):

```suggestion
            Assert.Equal("<module 'sys' (built-in)>", sys.ToString());
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace CSnakes.Runtime.Tests.Python;
+public class ImportTests : RuntimeTestBase
+{
+    [Fact]
+    public void TestImportModule()
+    {
+        using (GIL.Acquire())
+        {
+            using PyObject? sys = Import.ImportModule("sys");
```
atifaziz, line 12: The `?` is redundant since `Import.ImportModule` returns a non-nullable `PyObject`:

```suggestion
            using PyObject sys = Import.ImportModule("sys");
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.Equal("<module 'sys' (built-in)>", sys!.ToString());
+        }
+    }
+
+    [Fact]
+    public void TestReloadModule()
+    {
+        using (GIL.Acquire())
+        {
+            using PyObject? sys = Import.ImportModule("sys");
```
atifaziz, line 23:
```suggestion
            using PyObject sys = Import.ImportModule("sys");
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.NotNull(sys);
+            using PyObject? reloaded = Import.ReloadModule(sys!);
```
atifaziz, line 25:
```suggestion
            using PyObject reloaded = Import.ReloadModule(sys);
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.NotNull(reloaded);
+            Assert.Equal("<module 'sys' (built-in)>", reloaded!.ToString());
```
atifaziz, line 27:
```suggestion
            Assert.Equal("<module 'sys' (built-in)>", reloaded.ToString());
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    [Fact]
+    public void TestReloadModuleThatIsntAModule()
+    {
+        using (GIL.Acquire())
+        {
+            using PyObject? sys = PyObject.From<int>(42); // definitely not a module
```
atifaziz, line 36:
```suggestion
            using PyObject sys = PyObject.From<int>(42); // definitely not a module
```

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.NotNull(sys);
+            Assert.Throws<PythonInvocationException>(() => Import.ReloadModule(sys!));
```
atifaziz, line 38:
```suggestion
            Assert.Throws<PythonInvocationException>(() => Import.ReloadModule(sys));
```

atifaziz, line 38: Consider expanding the assertion to also the expected Python error type:

```suggestion
            var ex = Assert.Throws<PythonInvocationException>(() => Import.ReloadModule(sys));
            Assert.Equal("TypeError", ex.PythonExceptionType);
```

atifaziz (commented):
Really cool to have this, but I have some concerns about the design, potential memory leaks and exception safety that I've left as comments in the relevant parts.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -30,6 +30,9 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
 
             if (code is null) return;
 
+            // Calculate hash of code
+            var hash = code.ToString().GetHashCode();
```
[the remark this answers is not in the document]
atifaziz: I've submitted this suggestion as PR #261 against the the branch of this PR if you wish to apply it.

src/CSnakes.Runtime/CPython/Import.cs
```diff
@@ -39,5 +45,13 @@ protected static nint GetBuiltin(string name)
...
-    internal static partial nint PyImport_Import(nint name);
+    protected static partial nint PyImport_Import(nint name);
+
+    /// <summary>
+    /// Reload a module. Return a new reference to the reloaded module, or NULL with an exception set on failure (the module still exists in this case).
+    /// </summary>
+    /// <param name="module"></param>
+    /// <returns></returns>
+    [LibraryImport(PythonLibraryName)]
+    protected static partial nint PyImport_ReloadModule(PyObject module);
```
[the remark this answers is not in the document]
atifaziz: Something unrelated to this PR, but I am curious why these library imports send `PyObject` in but receive direct pointers via `nint`. I would have expected to see something along the lines of:

```c#
internal static partial PyObject PyImport_ReloadModule(PyObject module);
```

In fact, perhaps even something stronger:

```c#
internal static partial PyModule PyImport_ReloadModule(PyModule module);
```

src/Integration.Tests/python/test_changes.py
```diff
@@ -0,0 +1,3 @@
+
```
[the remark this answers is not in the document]
tonybaloney: this is an (odd) Python convention

src/CSnakes.Runtime/Python/Import.cs
```diff
@@ -8,4 +8,11 @@ public static PyObject ImportModule(string module)
     {
         return CPythonAPI.Import(module);
     }
+
+    public static PyObject ReloadModule(PyObject module)
```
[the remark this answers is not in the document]
tonybaloney: agreed, made those change

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
tonybaloney: It's a static class, so I'm not sure how you could implement this one simply?

atifaziz: See PR #269 as a demonstration of this.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
tonybaloney: > Won't this leak the previous function objects since the fields just get overwritten with potentially new objects?

The original object will get GC'ed and collected

> Is it exception-safe? What if the initialisation of the fields fails halfway? Won't it leave this in some strange state?

The exceptions are tied to the frame, which is associated to the code object. IF we still hold a reference to the old code object, we would need to close it.

> What happens if a function was removed, updated (the signature rather than the code) or added?

Hot Reload tells you that you need to restart (I tried this) and there are tests for that as well now

src/CSnakes.Runtime.Tests/Python/ImportTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿
+using CSnakes.Runtime.Python;
```
aaronpowell, line 2:
```suggestion
using CSnakes.Runtime.Python;
```

aaronpowell (approved):

PR #261, opened 2024-10-12
Use content hash for the module

atifaziz:
This PR addresses a [review suggestion](https://github.com/tonybaloney/CSnakes/pull/259#discussion_r1797696168) on PR #259.

### DIFF
--- a/src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
+++ b/src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
@@ -31,7 +31,7 @@ public class PythonStaticGenerator : IIncrementalGenerator
             if (code is null) return;
 
             // Calculate hash of code
-            var hash = code.ToString().GetHashCode();
+            var hash = code.GetContentHash();
 
             // Parse the Python file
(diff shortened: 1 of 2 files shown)


src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -150,4 +148,19 @@ public interface I{{pascalFileName}} : IDisposable
             }
             """;
     }
+
+    private static string HexString(ReadOnlySpan<byte> bytes)
```
tonybaloney, line 152: `Convert.ToHexString` should do this, no?

atifaziz: It's [only available since .NET 5](https://learn.microsoft.com/en-us/dotnet/api/system.convert.tohexstring), not for .NET Standard 2.0, which is what the source generator is targeting:

https://github.com/tonybaloney/CSnakes/blob/efd03f84eb3bac396d1bcd5408cdfc31b5b04333/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj#L3

tonybaloney (approved):

PR #263, opened 2024-10-12
Switched to the more performant System.Threading.Lock on .NET 9.0+

MarkCiliaVincenti:


AaronRobinsonMSFT:
I don't think this approach is needed for CSnakes. None of the new `Lock` features are needed here.

Just declaring a `class Lock;` for pre .NET 9 TFMs is more appropriate and doesn't introduce servicing issues with other libraries.

AaronRobinsonMSFT:
@MarkCiliaVincenti Would you be willing to update this PR to just go with the simpler approach below?

```csharp
#if NET9_0_OR_GREATER
global using Lock = System.Threading.Lock;
#else
global using Lock = object;
#endif
```

tonybaloney:
Thanks for the PR. Closing in favour of #289

MarkCiliaVincenti:
@tonybaloney  A new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

AaronRobinsonMSFT:
> @tonybaloney A new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

@MarkCiliaVincenti Something to consider with 3rd party source generators is that of servicing. By relying on a 3rd party library this library must monitor and trust that Backport.System.Threading.Lock will announce and manage servicing events. When one occurs this library then needs to respond and inform all its consumers. From my perspective consuming source generators that don't have a robust CVE process is a non-starter.

Eventhough CSnakes is an experimental library, it is being built to be consumed by products that have a fairly high degree of scrutiny on this front. I'm not sure it is appropriate to adopt any source generator outside of the built-in .NET product in this case.

PR #268, opened 2024-10-13
Make generated signature tests stricter

atifaziz:
This PR proposes to make the generated signature tests stricter by using `Assert.Equal` instead of `Assert.Contains`. Besides avoiding accidental matches, this change provides better assertion failures. With `Assert.Contains`, the error looks as follows:

    Xunit.Sdk.ContainsException
    Assert.Contains() Failure: Sub-string not found
    String:    "public void Hello(long? b = null)\r\n{\r\n   "···
    Not found: "void Hello(long? a = null)"

With `Assert.Equal`, one gets better information such as where the first difference occurs:

    Xunit.Sdk.EqualException
    Assert.Equal() Failure: Strings differ
                                       ↓ (pos 24)
    Expected: "public void Hello(long? a = null)"
    Actual:   "public void Hello(long? b = null)"
                                       ↑ (pos 24)

This helps in being able to spot and fix problems quicker.

### DIFF
--- a/src/CSnakes.Tests/GeneratedSignatureTests.cs
+++ b/src/CSnakes.Tests/GeneratedSignatureTests.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Logging;
 using CSnakes.Parser;
 using CSnakes.Reflection;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.Collections.Immutable;
 using System.ComponentModel;
 
 namespace CSnakes.Tests;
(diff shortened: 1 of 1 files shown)


aaronpowell (approved):

aaronpowell:
Looks like we got a conflict - give that a fix and should be good to go

atifaziz:
> Looks like we got a conflict - give that a fix and should be good to go

Resolved with dcf8b1b0f844118c3bb9fbcfc2785694a91b2527. Thanks, @aaronpowell.

PR #269, opened 2024-10-13
Have `ReloadModule` inherited/implemented through an interface

atifaziz:
This PR demonstrates an [idea being discussed](https://github.com/tonybaloney/CSnakes/pull/259#discussion_r1797715811) in PR #259.

### DIFF
--- a/src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
+++ b/src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
@@ -4,6 +4,7 @@ using CSnakes.Parser;
 using CSnakes.Parser.Types;
 using CSnakes.Reflection;
 using System.Collections.Immutable;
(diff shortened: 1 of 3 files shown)


src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -137,14 +135,16 @@ public void Dispose()
                     {{methods.Select(m => m.Syntax).Compile()}}
                 }
             }
-            public interface I{{pascalFileName}} : IDisposable
```
tonybaloney, line 140: IS the removal of `IDisposable` intentional?

atifaziz: It's inherited indirectly via `IReloadableModuleImport`.

tonybaloney (approved):

Issue #275, opened 2024-10-15
Make TagetFrameworks dependend on MSBuildVersion


minesworld:
net9.0 is only valid from MS VisualStudio 17.12 and upwards...

PR #276, opened 2024-10-15
fix for https://github.com/tonybaloney/CSnakes/issues/275

minesworld:
Make TagetFrameworks dependend on MSBuildVersion
net9.0 is only valid from MS VisualStudio 17.12 and upwards...


minesworld:
sorry commit to many files...

PR #278, opened 2024-10-16
Added WithExtraPaths() to EnvironmentBuilder

minesworld:
so those are changeable..


tonybaloney:
Thanks for the PR. 

If this is called multiple times, it shouldn't override the previous paths.

PR #287, opened 2024-10-20
Get error cause once

atifaziz:
This PR replaces two calls to get the `__cause__` attribute's value (`exception.GetAttr("__cause__")`) with a single one in order to avoid two hops across the marshalling boundary as well as allocations of the name.


tonybaloney (approved):

tonybaloney:
We probably need to extend the test cases for this as well

PR #288, opened 2024-10-20
Complete generator support for return value

atifaziz:
This PR completes the implementation of generator support by adding the testing the ability to retrieve the return value from a generator function.

Summary of changes:

- A new property `Return` is added to `IGeneratorIterator<,,>` that behaves like `IEnumerator<>.Current`.
- `Send` of `IGeneratorIterator<,,>` is changed to match semantics of `MoveNext`. As an aside, I would even propose to rename `Send` to `MoveNext` on the C# side since overloading the same name is more natural given that `next` in the Python is the same as `send(None)`.
- `Send` _implementation_ of `GeneratorIterator<,,>` is changed to:
  -  match semantics of `MoveNext`. The latter would return `false` whereas the former would throw `ArgumentOutOfRangeException` (with the wrong parameter for message) when the generator was exhausted.
  - retrieve the [`StopIteration.value`](https://docs.python.org/3/library/exceptions.html#StopIteration.value).
- `MoveNext` _implementation_ of `GeneratorIterator<,,>` is changed to simply delegate to the internal `Send` implementation with `PyObject.None` for value. This matches the Python semantics:
  > [New generator method: send(value)](https://peps.python.org/pep-0342/#new-generator-method-send-value)
  > Calling `send(None)` is exactly equivalent to calling a generator’s `next()` method.
- A new exception called `PythonStopIterationException` (that subclasses `PythonRuntimeException`) is introduced and which captures the `value` of `StopIteration` when it is raised so that it can be retrieved later. I don't think this is a great design on the .NET side, but this can be revisited later since the project is still a prototype. For now, it helps to complete generator support. Unfortunately, the implementation is also complicated by what I think is a regression in [`PyErr_Fetch`](https://docs.python.org/3.13/c-api/exceptions.html#c.PyErr_Fetch) that's been introduced with 3.12 ([see comment in code](https://github.com/tonybaloney/CSnakes/pull/288/files#diff-61bf667ff34807dbd5be0ae7e69313e020bee933fe6263e3eaafd6dc692834fbR11-R21)).

### DIFF
--- a/src/CSnakes.Runtime/IPythonEnvironment.cs
+++ b/src/CSnakes.Runtime/IPythonEnvironment.cs
@@ -5,6 +5,12 @@ namespace CSnakes.Runtime;
 
 public interface IPythonEnvironment : IDisposable
 {
+    /// <summary>
+    /// The version of the Python interpreter, which is a string that looks something like,
+    /// <c>"3.0a5+ (py3k:63103M, May 12 2008, 00:53:55) \n[GCC 4.2.3]"</c>. The first word (up to
+    /// the first space character) is the current Python version; the first characters are the major
+    /// and minor version separated by a period.
+    /// </summary>
     public string Version
     {
         get
--- a/src/CSnakes.Runtime/IPythonEnvironment.cs
+++ b/src/CSnakes.Runtime/IPythonEnvironment.cs
@@ -5,12 +5,6 @@ namespace CSnakes.Runtime;
 
 public interface IPythonEnvironment : IDisposable
 {
-    /// <summary>
-    /// The version of the Python interpreter, which is a string that looks something like,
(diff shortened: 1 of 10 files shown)


tonybaloney:
Please can you extend this test so that it verifies that the generator returned `True`, or change the generator to return something more exotic incase we accidentally return a truthy value anyway

https://github.com/tonybaloney/CSnakes/blob/main/src/Integration.Tests/GeneratorTests.cs#L14

atifaziz:
> Please can you extend this test so that it verifies that the generator returned `True`, …

I'm not sure what's missing since it was already extended.

aaronpowell (approved):

aaronpowell:
Looks good to me pending @tonybaloney's desires on testing

tonybaloney (approved):

tonybaloney:
Sorry I didn't see the additional test change. LGTM

PR #289, opened 2024-10-21
Use `System.Threading.Lock` on .NET 9.

AaronRobinsonMSFT:
Alternative proposal to https://github.com/tonybaloney/CSnakes/pull/263.

### DIFF
--- a/src/CSnakes.Runtime/CPython/Init.cs
+++ b/src/CSnakes.Runtime/CPython/Init.cs
@@ -12,7 +12,7 @@ internal unsafe partial class CPythonAPI : IDisposable
     public string PythonPath { get; internal set; } = string.Empty;
 
     private static string? pythonLibraryPath = null;
-    private static readonly object initLock = new();
+    private static readonly Lock initLock = new();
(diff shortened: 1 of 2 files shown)


MarkCiliaVincenti:
Works, but please refer to https://github.com/MarkCiliaVincenti/Backport.System.Threading.Lock?tab=readme-ov-file#why-not-keep-it-simple

aaronpowell:
I'm inclined to go with this approach rather than bringing in an external dependency like #263 does primarily because we're not likely to leverage the methods that the .NET 9 `Lock` type introduces - unless you can see areas in which we could benefit from it @MarkCiliaVincenti so I can better understand what limitations we might have in our current design.

MarkCiliaVincenti:
Nothing in particular, as long as you're aware. Should you need you can then easily switch to the version that uses the backport.

AaronRobinsonMSFT:
@aaronpowell or @tonybaloney Either of you want to sign off on this so we can get it in?

tonybaloney (approved):

PR #299, opened 2024-10-26
Fix optionality of function parameters & return

### LINKED ISSUE #295: Optional args/return either doesn't compile or throws at runtime
Suppose the following Python function:

```py
def test_optional_result(x: int) -> Optional[int]:
    return x if x >= 0 else None
```

I would have expected the C# method to return a `long?`, but instead it's typed to return just `long`:

```c#
long TestOptionalResult(long x);
```

This means that it's not possible to test for `null` _statically_, like so:

```c#
switch (testModule.TestOptionalResult(-1))
{
    case null:
        // ...
        break;
    case var result:
        // ...
        break;
}
```

since the compiler (rightfully) emits [CS0037](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0037), “Cannot convert null to 'long' because it is a non-nullable value type”. If one works around that by casting to `object` before testing:

```c#
//      V......V
switch ((object)testModule.TestOptionalResult(-1))
{
    case null:
        // ...
        break;
    case var result:
        // ...
        break;
}
```

the code compiles but results in the runtime exception:

    CSnakes.Runtime.PythonInvocationException
    Error converting Python object to int, check that the object was a Python int or that the value wasn't too large. See InnerException for details.
       at CSnakes.Runtime.CPython.CPythonAPI.PyLong_AsLongLong(PyObject p) in A:\CSnakes\main\src\CSnakes.Runtime\CPython\Long.cs:line 26

The inner exception reads:

    CSnakes.Runtime.PythonRuntimeException
    'NoneType' object cannot be interpreted as an integer
      Exception doesn't have a stacktrace

The same problem exists with input arguments that are typed as `Optional` of some type (when there's _no default_ of `None` given).

It looks like the problem is with [`TypeReflection.AsPredefinedType`](https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs#L17-L38), which simply looks at the type argument:

https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs#L33

The [existing nullability support](https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_defaults.approved.txt#L217-L220) is predicated on the presence of `None` as the _default argument_ in [`ArgumentReflection.ArgumentSyntax`](https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.SourceGeneration/Reflection/ArgumentReflection.cs#L12-L99):

https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.SourceGeneration/Reflection/ArgumentReflection.cs#L85-L88

The fix would be to consider `Optional` in the type hint as opposed to requiring a default of `None`. This is especially problematic for the return value.

atifaziz:
This PR fixes #295. It does this primarily by moving the decision of nullability to `TypeReflection.AsPredefinedType`, where the `Optional` type hint is considered. As a result of this, the `literalExpressionSyntax` in `ArgumentReflection.ArgumentSyntax` also becomes a simple switch expression that's solely based on the `PythonConstant` of the parameter's default value.

This also fixes some tests that were incorrect previously, like:

https://github.com/tonybaloney/CSnakes/blob/048e011b5d70849edb1d0c26fa9b985f8a48f6dc/src/CSnakes.Tests/GeneratedSignatureTests.cs#L30

The `b` parameter cannot have a default value of `None` per the type hint of `str`. The C# signature is correct, but because it was based on the presence of `None` as the default value.

`MethodReflection.ProcessMethodWithReturnType` needed some bigger changes because the return type's nullability needed to be handled with some extra care. `PyObject.As<T>` isn't good enough since it uses reflection based on type of `T` and isn't aware of nullability. As a result, `MethodReflection.ProcessMethodWithReturnType` generates code to check for `None` and return `null` instead when the return type is `Optional`. I've added  integration tests for this (see `test_optional.py`), where you can see this in action along the lines of:

```c#
public long? TestInt(long? n)
{
    using (GIL.Acquire())
    {
        // ...
        return __result_pyObject.IsNone() ? null : __result_pyObject.As<long>();
    }
}
```


tonybaloney:
> The b parameter cannot have a default value of None per the type hint of str. 

I added that test because type hints are just hints. Python developers often (although it is wrong) use default of `None` with a parameter that is not marked as optional either via `Optional[str]` or `Union[str, None]` or `str | None`

tonybaloney (approved):

PR #308, opened 2024-11-05
Fix pythonpath (again)

minesworld:
This time it should work...

- added test for __file__ and __name__ attribute on os module (first commit, so should fail without the patch)

- removed PySet_Path usage and DLL definitions
- added a AppendMissingPathsToSysPath() to CPythonAPI, called within GIL before initializing the types
- AppendMissingPathsToSysPath does as it says: it checks if a path is already in sys.path, if not it will be appended
- PythonPath is splitted "as is", its up to the setter of PythonPath to provide absolute paths

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/AttributeTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/AttributeTests.cs
@@ -1,0 +1,25 @@
+﻿using CSnakes.Runtime.Python;
+namespace CSnakes.Runtime.Tests.Python;
+public class AttributeTests : RuntimeTestBase
+{
+    [Fact]
+    public void TestHasAttrFile()
+    {
+        using (GIL.Acquire())
+        {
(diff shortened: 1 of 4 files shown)


minesworld:
@Tony : my first pull request failed to work as it was based on the wrong fork. The first thing I did was refactoring the CAPI...

```
PyList_GetItem
PyList_GetItem_
```

and an (now) an additional  

`PyList_GetItemRaw`

Sorry - but WTF ?!?

My "minesworld" fork refactoring the raw calls into an "Unmanaged" nint based CAPI which is the base for the normal IntPtr CAPI is much "cleaner"... https://github.com/minesworld/CSnakes/tree/minesworld

But the **real beauty** is the kind of genious `NewReference `/ `BorrowedReference `C# language usage of pythonnet.

The only downside to that might be that there is no raw nint and such it might have effects on the heap/stack and is a bit slower...

PR #313, opened 2024-11-14
Add support for async functions

tonybaloney:
Concept:

- A `Coroutine[T_Yield, T_Send, T_Return]` will be mapped to an `ICoroutine<Tyield, TSend, TReturn>`, similar to Generator
- The Coroutine class has an implementation for `AsTask()` which returns a `Task<TYield>` by executing the following:
 - Creating a Python event loop if one doesn't exist. Event loops are 1:1 with threads, since .NET reuses threads for tasks from a thread pool, the event loops are reused using thread local statics. All of the event loops are closed at shutdown
 - Generating a task from the Python coroutine object
 - Scheduling the task
- All of this is done within a thread in .NET Task

The code generator has been updated to reflect the function to `Task<TYield>` as moves the `AsTask()` execution outside of the initial GIL block to ensure they don't get deadlocked.


tonybaloney:
Tests are hanging because `Py_Finalize` is called from a different thread than `Py_Initialize`. See https://github.com/python/cpython/issues/122517

src/CSnakes.Runtime/CPython/Init.cs
```diff
@@ -65,7 +67,47 @@ internal void Initialize()
...
+                    {
+                        Thread.Sleep(500);
+                    }
+                }
+            },
+            cancellationToken: cts.Token);
+
+        while (!initialized) // Wait for startup
+        {
+            continue;
```
tonybaloney, line 109: I'm sure there's a better pattern for doing this, also we should maybe add a 50ms sleep here

src/CSnakes.Runtime/CPython/Init.cs
```diff
@@ -126,28 +170,48 @@ private void InitializeEmbeddedPython()
...
+            if (!IsInitialized)
+                return;
+
+            // Shut down asyncio coroutines
+            CloseEventLoops();
+
+            // Clean-up interns
+            NewEventLoopFactory?.Dispose();
+            AsyncioModule?.Dispose();
+            // TODO: Add more cleanup code here
```
tonybaloney, line 186: This TODO is to clean up the currently static globals which we never dispose of, but are cleaned up automatically. We should dispose of them incase people recycle environments.

src/CSnakes.Runtime/CPython/Coroutine.cs
```diff
@@ -0,0 +1,82 @@
...
+        return HasAttr(p, "__await__");
+    }
+
+    internal class EventLoop : IDisposable
+    {
+        private readonly PyObject? loop = null;
+
+        public EventLoop()
+        {
+            loop = NewEventLoopFactory!.Call();
```
AaronRobinsonMSFT, line 19: This call can return a `null`? If so, why not simply throw here if `null`?

src/CSnakes.Runtime/CPython/Coroutine.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            if (loop is null)
+            {
+                throw new InvalidOperationException("Event loop not initialized");
+            }
+            using var close = loop.GetAttr("close");
+            close?.Call();
+        }
+
+        public void Dispose()
```
AaronRobinsonMSFT, line 34: We should ensure that `Dispose()` can be called multiple times. Not sure if this implementation handles that.

src/CSnakes.Runtime/CPython/Coroutine.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            Close();
+            loop?.Dispose();
+            IsDisposed = true;
+        }
+
+        public PyObject RunTaskUntilComplete(PyObject coroutine)
```
AaronRobinsonMSFT, line 41: Public APIs on a type that implement `IDisposable` typically throw an `ObjectDisposedException` when they are disposed.

src/CSnakes.Runtime/CPython/Init.cs
AaronRobinsonMSFT: Perhaps [`TaskCompletionSource<T>`](https://learn.microsoft.com/dotnet/api/system.threading.tasks.taskcompletionsource-1)?

AaronRobinsonMSFT: Could also use some event (for example, `ManualResetEventSlim` or `AutoResetEvent`) instead of a loop in the above initialization thread.

richardu:
Hi @tonybaloney (et. al),
Thanks for this great tool!
RE: async functionality - what's the current status of it? alpha / beta / RC, etc.?
Thx, Rich

src/CSnakes.Runtime/CPython/Init.cs
tonybaloney: ManualResetEventSlim is a good mechanism

PR #320, opened 2025-01-05
Improved handling of name generation with single-character segments

jozefhornik:
Single-character segments with underscores were previously ignored during processing. This caused errors when argument names like test_x and test_y were both transformed into test_.

The primary fix is implemented in the CSnakes.CaseHelper.ToPascalCase method, along with updates to the corresponding tests.


tonybaloney (approved):

tonybaloney:
Ah. Good catch, thanks for the fix

PR #323, opened 2025-01-09
Fetch a precompiled build of Python

### LINKED ISSUE #324: Fetch a precompiled version of Python instead of depending on one in the system
# Feature: 

To bundled Python packages and ship them as nuget packages, it is unreasonable to expect the end user to have or know where Python is installed.

Use the new python-static-builds to have a simple zero-arguments locator which downloads and installs Python into a temporary folder and uses that for everything in CSnakes.

Related #322

tonybaloney:
Adds a new locator that downloads a precompiled build of Python and uses that.

Means that users don't need to have Python setup before using a package.

Fixes #324


src/CSnakes.Runtime/ServiceCollectionExtensions.cs
```diff
@@ -179,6 +179,18 @@ public static IPythonEnvironmentBuilder FromConda(this IPythonEnvironmentBuilder
         return builder;
     }
 
+    public static IPythonEnvironmentBuilder FromManagedPython(this IPythonEnvironmentBuilder builder)
```
tonybaloney, line 182: Not wedded to this name. Suggestions welcome

tonybaloney:
Failing because I'm creating symlinks when the entire folder might not exist

src/CSnakes.Runtime/Locators/ManagedPythonLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            return Path.Combine(folder, "bin", $"python{Version.Major}.{Version.Minor}{suffix}");
+        }
+
+        throw new PlatformNotSupportedException($"Unsupported platform: '{RuntimeInformation.OSDescription}'.");
+    }
+
+    public override PythonLocationMetadata LocatePython()
+    {
+        var downloadPath = Path.Join(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "CSnakes", $"python{Version.Major}.{Version.Minor}");
```
aaronpowell, line 37: AppData or Temp?

src/CSnakes.Runtime/Locators/ManagedPythonLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+            else
+            {
+                throw new PlatformNotSupportedException($"Unsupported platform: '{RuntimeInformation.OSDescription}'.");
+            }
+            string downloadUrl = $"https://github.com/astral-sh/python-build-standalone/releases/download/{standaloneRelease}/cpython-{Version.Major}.{Version.Minor}.{Version.Build}+{standaloneRelease}-{platform}.tar.zst";
+
+            // Download and extract the Zstd tarball
+            logger.LogInformation("Downloading Python from {DownloadUrl}", downloadUrl);
+            string tempFilePath = DownloadFileToTempDirectoryAsync(downloadUrl).GetAwaiter().GetResult();
```
aaronpowell, line 109: Should we refactor the base method to be async?

src/CSnakes.Runtime/Locators/ManagedPythonLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+
+    private static string DecompressZstFile(string zstFilePath)
+    {
+        string tarFilePath = Path.ChangeExtension(zstFilePath, ".tar");
+        using (var inputStream = new FileStream(zstFilePath, FileMode.Open, FileAccess.Read))
+        using (var decompressor = new DecompressionStream(inputStream))
+        using (var outputStream = new FileStream(tarFilePath, FileMode.Create, FileAccess.Write))
+        {
+            decompressor.CopyTo(outputStream);
+        }
```
aaronpowell, line 179:
```suggestion
        using var inputStream = new FileStream(zstFilePath, FileMode.Open, FileAccess.Read);
        using var decompressor = new DecompressionStream(inputStream);
        using var outputStream = new FileStream(tarFilePath, FileMode.Create, FileAccess.Write);
        decompressor.CopyTo(outputStream);
```

Just so it's consistent with how other `using`'s are done.

src/CSnakes.Runtime/Locators/ManagedPythonLocator.cs
```diff
@@ -0,0 +1,215 @@
...
+        return tarFilePath;
+    }
+
+    private static void ExtractTar(string tarFilePath, string extractPath, ILogger logger)
+    {
+        using FileStream tarStream = File.OpenRead(tarFilePath);
+        using TarReader tarReader = new(tarStream);
+        TarEntry? entry;
+        List<(string, string)> symlinks = [];
+        while ((entry = tarReader.GetNextEntry()) != null)
```
aaronpowell, line 189:
```suggestion
        while ((entry = tarReader.GetNextEntry()) is not null)
```

src/CSnakes.Runtime/ServiceCollectionExtensions.cs
aaronpowell: `FromPythonDownload`?

`FromPythonRedistributable`?

src/CSnakes.Runtime/ServiceCollectionExtensions.cs
```diff
@@ -179,6 +179,18 @@ public static IPythonEnvironmentBuilder FromConda(this IPythonEnvironmentBuilder
...
+    {
+        builder.Services.AddSingleton<PythonLocator>(
+            sp =>
+            {
+                var logger = sp.GetRequiredService<ILogger<ManagedPythonLocator>>();
+                return new ManagedPythonLocator(logger);
+            }
+        );
```
aaronpowell, line 190:
```suggestion
        builder.Services.AddSingleton<PythonLocator, ManagedPythonLocator>();
```

Registers the type `ManagedPythonLocator` to be resolved as `PythonLocator`.

aaronpowell (changes requested):

src/CSnakes.Runtime/Locators/ManagedPythonLocator.cs
tonybaloney: AppData. It uses temp for the tar.zst file then deletes it. Multiple apps using csnakes should use the same downloaded copy

src/CSnakes.Runtime/ServiceCollectionExtensions.cs
tonybaloney: `FromRedistributable` makes sense since none of the others have Python in the name

PR #330, opened 2025-01-12
Use system mutex in `RedistributableLocator` during installation

atifaziz:
This PR suggests to replace installation lock file with the use of a named [`Mutex`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-9.0#system-threading-mutex-ctor(system-boolean-system-string)), which requires less code (no polling with sleeps and lock file management), supports timing out, has cross-platform support and doesn't suffer from race conditions of file operations (which require more care). Moreover, a process that crashes and whose thread owned the mutex can signal [abandonment](https://github.com/dotnet/runtime/blob/9d5a6a9aa463d6d10b0b0ba6d5982cc82f363dc3/src/coreclr/pal/src/include/pal/mutex.hpp#L111-L116) to others so they can acquire the lock.

Incidentally, the [`Mutex` implementation on some non-Windows platforms also uses a lock file](https://github.com/dotnet/runtime/blob/9d5a6a9aa463d6d10b0b0ba6d5982cc82f363dc3/src/coreclr/pal/src/include/pal/mutex.hpp#L94-L98) behind the scenes.

### DIFF
--- a/src/CSnakes.Runtime/Locators/RedistributableLocator.cs
+++ b/src/CSnakes.Runtime/Locators/RedistributableLocator.cs
@@ -8,6 +8,7 @@ namespace CSnakes.Runtime.Locators;
 internal class RedistributableLocator(ILogger<RedistributableLocator> logger, int installerTimeout = 360) : PythonLocator
 {
     private const string standaloneRelease = "20250106";
+    private const string MutexName = @"Global\CSnakesPythonInstall-1"; // run-time name includes Python version
     private static readonly Version defaultVersion = new(3, 12, 8, 0);
     protected override Version Version { get; } = defaultVersion;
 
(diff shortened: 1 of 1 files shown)


tonybaloney:
The file is so that multiple processes can work together. The CI failures are probably showing that failure because they're running on the same image and trying to write over each other.

I did look at an OS semaphore as a sync primitive but didn't know how cross-platform that is for .NET

tonybaloney (approved):

atifaziz:
> The file is so that multiple processes can work together.

Right, and thus globally named mutex.

> The CI failures are probably showing that failure because they're running on the same image and trying to write over each other.

I'm not sure.

> I did look at an OS semaphore as a sync primitive but didn't know how cross-platform that is for .NET

The [design is cross-platform, as follows](https://github.com/dotnet/runtime/blob/9d5a6a9aa463d6d10b0b0ba6d5982cc82f363dc3/src/coreclr/pal/src/include/pal/mutex.hpp#L72-L76) (though details may change in the future as the techniques used are not officially documented):

> Design
>
> - On systems that support pthread process-shared robust recursive mutexes, they will be used
> - On other systems, file locks are used. File locks unfortunately don't have a timeout in the blocking wait call, and I didn't
  find any other sync object with a timed wait with the necessary properties, so polling is done for timed waits.

I think the real problem was that I forgot to release the mutex explicitly since its disposal doesn't do that. I've done that with cea7bd4dde2ad8956fa3ea0a306ae09ef20c446b and the CI tests all pass. However, this means the abandonment code may not be working as imagined. Since I wasn't releasing the mutex before, each waiting install would have considered the other as abandoned, _in turn_, and deleted the directory and restarted. It would have been super inefficient, but shouldn't have lead to failures and this still bothers me and I'll keep thinking about it.

PR #340, opened 2025-01-15
Allow users to run Python code from a string

tonybaloney:
Implements #290 

Adds two different types of API:
- IPythonEnvironment.ExecuteExpression which takes a simple expression (`1+1`) and returns the result
- IPythonEnvironment.Execute which takes a string, typically multiple lines

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/RunTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/RunTests.cs
@@ -6,14 +6,14 @@ public class RunTests : RuntimeTestBase
     [Fact]
     public void TestSimpleString()
     {
-        using var result = env.Execute("1+1");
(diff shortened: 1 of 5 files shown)


aaronpowell (approved):
While the code itself looks fine I question the general value of this, as what role is .NET really playing here? Like, you really need to know the inputs and outputs beforehand so that you can read/write them, or you just pass through arbitrary code with .NET not really "touching" anything at either end (which seems inherently dangerous)

tonybaloney:
This would be used for edge cases, like running complicated expressions on an object that you've got back from CSnakes, or dynamically running a bit of interpreted code on the module.

aaronpowell:
Yeah, I read the issue so I "get" the use-case but it's the kind of thing I wonder if CSnakes is the best tool for, since .NET isn't exactly idea for a scripting environment.

PR #369, opened 2025-03-13
Fix crash on GIL acquisition within pybind11 extensions

### LINKED ISSUE #357: v1.0.26 and greater throws System.ExcecutionEngineException
I have been running the following python for a while with no problems until upgraded to **v1.0.26** and higher which now throws at the following line:

`__underlyingPythonFunc.Call`

### Python code:

```
import gc
from typing import Generator
from fastembed import TextEmbedding

embedding_model = None

def initialize(model_name: str = "BAAI/bge-small-en-v1.5") -> None:
    global embedding_model
    if embedding_model is not None:
        del embedding_model
        gc.collect()

    embedding_model = TextEmbedding(model_name)
    print(f"The model '{model_name}' is ready to use.")

def generate_query_embedding(query: str) -> list[float]:
    if embedding_model is None:
        raise RuntimeError("Embedding model is not initialized. Call 'initialize' first.")

    return list(embedding_model.query_embed(query))[0]
```

### Generated code:

```csharp
public IReadOnlyList<double> GenerateQueryEmbedding(string query)
{
    using (GIL.Acquire())
    {
        logger.LogDebug("Invoking Python function: {FunctionName}", "generate_query_embedding");
        PyObject __underlyingPythonFunc = this.__func_generate_query_embedding;
        using PyObject query_pyObject = PyObject.From(query)!;
        using PyObject __result_pyObject = __underlyingPythonFunc.Call(query_pyObject);
        return __result_pyObject.As<IReadOnlyList<double>>();
    }
}
```

### Setup:

```csharp
services
    .WithPython()
    .FromNuGet("3.12.7")
    .WithVirtualEnvironment(Path.Join(home, ".venv"))
    .WithPipInstaller(requirements.FullName)
```

### Requirements
```
pandas
fastembed
```

### LINKED ISSUE #362: When I try to run Paddle OCR, it is likely to show "Process finished with exit code 139."
When I use the OCR capability of paddle, the program always shows "Process finished with exit code 139."

There is a small chance of success once, with an element of chance. However, executing scripts with Python commands is very stable and will not result in exceptions.

I am not sure whether this problem is caused by CSnakes or other reasons.

Tried different versions of Python 3.12.9 / 3.10.16 / 3.9.21, none of them worked.

In order to minimize the reproduction of the model, the 3.12.9 version of CSnakes provided by this Issues is used.

Need to manually install the paddlepaddle package and paddleocr in the .venv 

Paddle documentation
https://paddlepaddle.github.io/PaddleX/latest/installation/paddlepaddle_install.html

```
python -m pip install paddlepaddle==3.0.0rc1 -i https://www.paddlepaddle.org.cn/packages/stable/cpu/
```

```
pip install paddleocr
```

```python
from paddleocr import PaddleOCR,draw_ocr

def predict() -> None:
    ocr = PaddleOCR()
    img_path = './images/test.jpeg'
    result = ocr.ocr(img_path,rec=False)
    for idx in range(len(result)):
        res = result[idx]
        for line in res:
            print(line)
    
    from PIL import Image
    result = result[0]
    image = Image.open(img_path).convert('RGB')
    im_show = draw_ocr(image, result, txts=None, scores=None, font_path='./simfang.ttf')
    im_show = Image.fromarray(im_show)
    im_show.save('./result.jpg')

```

```c#
var appFolder = Path.GetDirectoryName(Environment.GetCommandLineArgs()[0]);

var builder = Host.CreateDefaultBuilder(args)
    .ConfigureServices(services =>
    {
        services
            .WithPython()
            .WithHome(appFolder)
            .WithVirtualEnvironment(Path.Combine(appFolder, ".venv"))
            .FromRedistributable();
    });
var app = builder.Build();
var env = app.Services.GetRequiredService<IPythonEnvironment>();
var paddleOcr = env.PaddleOcr();
paddleOcr.Predict();
```

Logs ⬇️

```
dbug: CSnakes.Runtime.EnvironmentManagement.VenvEnvironmentManagement[0]
      Adding environment site-packages to extra paths: /Users/_/Documents/GitHub/example-csnakes/bin/Debug/net9.0/.venv/lib/python3.12/site-packages
dbug: CSnakes.Runtime.EnvironmentManagement.VenvEnvironmentManagement[0]
      Virtual environment already exists at /Users/_/Documents/GitHub/example-csnakes/bin/Debug/net9.0/.venv
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Setting up Python environment from /Users/_/Library/Application Support/CSnakes/python3.12.9/python/install using home of /Users/_/Documents/GitHub/example-csnakes/bin/Debug/net9.0
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Python DLL: /Users/_/Library/Application Support/CSnakes/python3.12.9/python/install/lib/libpython3.12.dylib
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Python path: /Users/_/Library/Application Support/CSnakes/python3.12.9/python/install/lib/python3.12:/Users/_/Library/Application Support/CSnakes/python3.12.9/python/install/lib/python3.12/lib-dynload
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Adding extra paths to PYTHONPATH: /Users/_/Documents/GitHub/example-csnakes/bin/Debug/net9.0/.venv/lib/python3.12/site-packages
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Importing module paddle_ocr
dbug: CSnakes.Runtime.IPythonEnvironment[0]
      Invoking Python function: predict

Process finished with exit code 139.
```

tonybaloney:
Fixes #357 
Fixes #362


src/CSnakes.Runtime/Python/GIL.cs
```diff
@@ -26,25 +26,33 @@ namespace CSnakes.Runtime.Python;
...
+            }
             Debug.Assert(CPythonAPI.IsInitialized);
-            gilState = CPythonAPI.PyGILState_Ensure();
+            CPythonAPI.PyEval_RestoreThread(pythonThreadState);
             recursionCount = 1;
         }
 
         public void Ensure()
         {
             if (recursionCount == 0)
```
ericsnowcurrently, line 53: In the alternate branch, we should expect there is a current thread state, right?  Would it be worth adding an assert here to that effect?

Would we also expect that it would always be `pythonThreadState`?

src/CSnakes.Runtime/Python/GIL.cs
```diff
@@ -62,7 +70,7 @@ public void Dispose()
             {
                 CPythonAPI.Py_DecRefRaw(handle);
             }
-            CPythonAPI.PyGILState_Release(gilState);
+            pythonThreadState = CPythonAPI.PyEval_SaveThread();
```
ericsnowcurrently, line 73: The tracked thread state should be the returned one, right?  It may be worth verifying that's the case, e.g. with an assert.

src/CSnakes.Runtime/Python/GIL.cs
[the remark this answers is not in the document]
atifaziz: @ericsnowcurrently Below are AI-generated 🤖 answers to your questions that have been validated by me, a human 😅. I've added my own comments in _italics_.

1. **Is a public method of an internal class visible to users?**
   - In C#, a public method of an internal class is not visible to users outside the [assembly]. The `internal` keyword restricts the visibility of the class to the assembly (_think DLL or library as a gross simplification_) in which it is declared.

2. **Can instances of an internal class be stored by users (e.g. assigned to a variable)?**
   - Instances of an internal class cannot be stored by users outside the assembly. The `internal` keyword restricts the visibility and use of the class to the same assembly.

3. **Is `IDisposable` the protocol that facilitates `using` statements, and thus the equivalent of the context manager protocol in Python?**
   - Yes, `IDisposable` is the interface that facilitates `using` statements in C#, similar to the context manager protocol in Python _used in conjunction with `with`_.

4. **Is `Dispose()` the equivalent of `__exit__()` in Python?**
   - Yes, the `Dispose()` method in C# is equivalent to the `__exit__()` method in Python's context management protocol.

5. **Is `Dispose()` also a finalizer of sorts?**
   - No, `Dispose()` is not a finalizer. `Dispose()` is used for deterministic resource cleanup, while finalizers (`~ClassName()`) are used by the garbage collector for non-deterministic cleanup (_like `__del__` in Python when GC is involved in breaking cycles_).

6. **From a language/syntax standpoint, can you require `GIL.Acquire()` only be used in a `using` statement?**
   - No, _like Python_, C# does not enforce that a method can only be used in a `using` statement. However, it is recommended to use `IDisposable` objects within a `using` statement for proper resource management. _You can use [static code analyzers to somewhat enforce this](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213)_.

7. **Can a user call `GIL.Acquire()` without putting it in a `using` statement or chaining a method call (e.g. `GIL.Acquire().Dispose()`)?**
   - Yes, a user can call `GIL.Acquire()` without using a `using` statement or chaining a method call, but it is not recommended. The `using` statement ensures proper disposal of resources.

8. **How does C# know how to clean up `GIL` instances?**
   - C# relies on the `IDisposable` interface and the `Dispose()` method for cleaning up `GIL` instances. Additionally, finalizers can be used for non-deterministic cleanup.
   - _It's not that C# relies on `IDisposable` per se, but that [it's a pattern](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern) that's up to the developer to use when deterministic cleanup is critical to proper operation._

9. **When does that happen?**
   - The cleanup happens when the `Dispose()` method is called explicitly or implicitly by the `using` statement. Finalizers are called by the garbage collector when the object is no longer reachable.

10. **How does it know how to clean up `GIL.pyThreadState` (or `GIL.currentstate` for that matter)?**
    - The `Dispose()` method in the `GIL` class is responsible for cleaning up the `GIL.pyThreadState` and `GIL.currentstate` by releasing the GIL and disposing of any remaining handles.

11. **Does being a static class mean that `GIL` is essentially a singleton instance?**
    - Yes, being a static class means that `GIL` is a singleton in the sense that it cannot be instantiated, and all its members are shared across the application. _By some stretch of imagination, it's almost akin to a module in Python_.

[assembly]: https://learn.microsoft.com/en-us/dotnet/standard/assembly/

PR #372, opened 2025-03-15
Make GIL disposal idempotent once disposed

atifaziz:
If `GIL.Disposed` is called more times than `GIL.Acquire` then a fatal error is reported

    Exit code is -1073740791 (Fatal Python error: PyEval_SaveThread: the function must be called with the GIL held, after Python initialization and before Python finalization, but the GIL is released (the current Python thread state is NULL)
        Python runtime state: initialized
        
        Current thread 0x000088dc (most recent call first):
          <no Python frame>
        
        Thread 0x000088cc (most recent call first):
          <no Python frame>)

This PR fixes the problem by making `GIL.Dispose` idempotent once the recursion count falls to 0 and the GIL has been previously released.

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/GILTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/GILTests.cs
@@ -1,0 +1,13 @@
+using CSnakes.Runtime.Python;
+
+namespace CSnakes.Runtime.Tests.Python;
+public class GILTests : RuntimeTestBase
+{
+    [Fact]
+    public void DisposeIsIdempotentOnceDisposed()
+    {
+        var gil = GIL.Acquire();
+        gil.Dispose();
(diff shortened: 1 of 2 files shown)


tonybaloney (approved):

tonybaloney:
Good spot. We need more tests for the GIL wrapper. 

This pybind11 bug was a nightmare to solve

atifaziz:
> We need more tests for the GIL wrapper.

Add some basic ones with PR #373.

atifaziz:
> Good spot.

Credit goes to @ericsnowcurrently whose [questions](https://github.com/tonybaloney/CSnakes/pull/369#discussion_r1996107214) prompted me to look a bit deeper and ensure the balance of GIL locks couldn't go overdraft.

PR #375, opened 2025-03-20
Cleanup dependencies now that .NET 9 has hit RTM

buvinghausen:
Tony thank you for the excellent talk today with Hanselman I am in love with this library!

I noticed when digging through my dependency graph that it was still pulling in RC2 .NET 9 stuff as well as System.Text.Json but STJ is included in the runtime when your target framework moniker is set at 8 or 9.


tonybaloney:
Thanks for the PR! Well spotted.

tonybaloney (approved):

Issue #379, opened 2025-03-21
Silent Restart of .NET App When Invoking Python Code That Uses CUDA (via PyTorch)


chaseaucoin:
Hello,

I recently came across CSnakes through your conversation with Scott Hanselman on YouTube — really impressive work! I’m experimenting with integrating it into a text-to-voice system that uses PyTorch and Kokoro, and it seems like a great fit for the project.

However, I ran into an unexpected behavior that I’m having trouble diagnosing. Specifically, when the code reaches the line where `KPipeline` is instantiated (`PipelineManager._pipeline = KPipeline(lang_code='a')`), there’s no error or exception thrown, but the entire .NET program appears to restart. Execution jumps all the way back to the start of `Program.cs`, right at the `"Creating the builder..."` log message.

This caught me off guard — it felt like a silent crash or process restart, rather than a Python exception. It took a while to realize that this was happening consistently at that exact point.

To isolate the issue, I initially had the pipeline instantiation at the root of the module, but moved it inside a class (`PipelineManager`) to see if that made any difference — it didn’t.

There’s no visible error output or traceback, so I’m unsure if this is a problem with how CSnakes is invoking or managing Python code at runtime, or if something lower level is crashing silently (perhaps related to native extensions or the CPython interaction?).

Do you have any insight into what might cause this kind of behavior, or how I could further debug it? I’d be happy to share more details or set up a minimal repro if needed.

python/requirements.txt
```
--find-links https://download.pytorch.org/whl/torch_stable.html
--extra-index-url https://download.pytorch.org/whl/cu126
torch==2.6.0
torchaudio==2.6.0
torchvision==0.21.0
kokoro==0.8.4
```

python/speech_service.py

```python
from kokoro import KPipeline
import torch
import time

class PipelineManager:
    _pipeline = None
    _model = None

    @staticmethod
    def get_pipeline():
        if PipelineManager._pipeline is None:
            print("Loading pipeline")
            # Wrap in try except to catch errors then print them
            try:
                PipelineManager._pipeline = KPipeline(lang_code='a')
            except Exception as e:
                print("Error loading pipeline")
                print(e)
            print("Loading voice")
            PipelineManager._pipeline.load_voice("af_heart")
        return PipelineManager._pipeline

    @staticmethod
    def get_model():
        if PipelineManager._model is None:
            print("Loading model")
            PipelineManager._model = PipelineManager.get_pipeline().model
        return PipelineManager._model

speed = 1


[...]

PR #380, opened 2025-03-25
Set sys.executable on startup to the python executable path

tonybaloney:
A permanent fix for the behaviour reported in #379


tonybaloney:
There are 2 other ways to do this:

1. Use `Py_SetProgramName` which has been deprecated since 3.11 and is probably going to be removed in 3.14
2. Change the whole initialization mechanism to PyConfig pointers. Which I don't want to do for this small change

aaronpowell (approved):

PR #383, opened 2025-03-29
Allow run Conda tests with Miniconda3

gabrielgt:
Check if the anaconda3 directory exists and if not, change it to miniconda3


tonybaloney:
Thanks for that. 

Why is the diff on this PR so huge? I think it needs to be rebased.

PR #386, opened 2025-03-31
Use PolySharp package for poly-filling

atifaziz:
This PR removes the need for the local pollyfill `NotNullAttributePolyfill` by using the [PolySharp](https://www.nuget.org/packages/PolySharp) package instead. PolySharp embeds the pollyfills by generating them into the consuming project. Moreover, PolySharp comes with several other pollyfills that are being explicitly enabled here:

- `System.Index`
- `System.Range`
- `System.Diagnostics.CodeAnalysis.NotNullAttribute`
- `System.Diagnostics.CodeAnalysis.NotNullWhenAttribute` (needed by `System.Index`)
- `System.Diagnostics.CodeAnalysis.DoesNotReturnAttribute` (needed by `System.Index`)

Adding `System.Index` helps to simplify some code by being able to take advantage of list patterns in pattern matching. This can be seen in `MethodReflection` where the `Coroutine` type specification is validated and its type argument tested for `None`. The same happens in `TypeReflection` where patterns can be used to test for generic types and capture the type arguments via list patterns (implicitly checking/matching the length too).


tonybaloney:
Looks cleaner to me. @aaronpowell ?

Issue #314, opened 2024-11-22
Improve the first 30-minutes experience


tonybaloney:
We need to improve the developer experience for the first 30 minutes.

1. The Host Builder assumes a DI container- what happens if they don't have one? What is the simple pattern for a Console Application?
2. You need 3 locators to be multi-platform. Linux and MacOS are hard because you can't just locate then binaries (see related issues). 
3. Specifying the "home" path isn't obvious when it's a relative path to the source files. 

If anyone else has any thoughts on how things could be easier/simpler to get started please add in the comments.

NimaAra:
I think the overall experience is good, all I needed was to go through the docs and was up and running in a few minutes.

Here are my 2cents if you need it:

* Requiring a DI container is fine but not needing one is even "finer"
* In my case (and it's specific to my scenario) I care less about the source generated workflow and more about hand written integration.
* I do agree with having to worry about locators, they seem confusing, I would have prefered the `.WithPython()` to have an option/setting on which I can specify where to find _Python_ rather than using the _FromNuget/FromEnvVar_ etc. Allowing these options on the option/setting would be simpler IMO.
* It is not clear to me why I need `.WithHome` when I am already providing the paths to the _.py_ files, do all my referenced _.py_ files need to live in the module path?

jmichas:
Well, I don't want to sound like jerk but, I think the first 30m experience is horrible. I'm in more like the first 3h and I still can't get it working.

I am working on MacOS using Rider.

I agree with @NimaAra about using an option pattern with WithPython(). The other extension methods do things that aren't immediately apparent ie appending things to the path supplied.

I am surprised that the FromNuGet() method only supports windows. I tried to create my own to support MacOS but because downstream it appends "bin/xyz" type stuff it doesn't work or requires more work from me overriding all the GetXXX methods in the Locator.

I then tried to just setup my own environment variable but that too gets appended with other path stuff so it fails.

My goal is to provide a .Net MAUI app that wraps some customized Python libraries, but I don't see how I would provide this where Python gets installed with the MAUI app.

Anyway my current tests are just with a simple console app.

NimaAra:
@jmichas, You do have some control over setting the various required paths. I used the following, you need to set the paths according to your python installation. I suggest you use [uv](https://github.com/astral-sh/uv) to manage your python installations.

```csharp
internal sealed class ContainerLocator : PythonLocator
{
[...]

PR #387, opened 2025-04-02
Use redistributable Python for runtime- and integration tests

gukoff:
This PR makes these unit tests runnable on all platforms out of the box.

For me, running unit tests on OSX was failing because I never installed Python with an official installer, the way `MacOSInstallerLocator` expects it.

Perhaps a better unit test setup would be to always use redistributable for general purpose unit tests, plus separate test suites for the specific Python locators, like it's now done for Conda.

---

Another potential improvement would be to return `null` from the locator if it can't find a Python installation. This seems to be the expected behavior - see the last line of the LINQ below from `PythonEnvironment`. However, `LocatePythonInternal()` throws exceptions when it can't find Python.

```cs
var location = locators
    .Where(locator => locator.IsSupported())
    .Select(locator => locator.LocatePython())
    .FirstOrDefault(loc => loc is not null);
```

Related: #314


tonybaloney:
>Perhaps a better unit test setup would be to always use redistributable for general purpose unit tests, plus separate test suites for the specific Python locators, like it's now done for Conda.

Great idea. We introduced the redistributable locator quite recently, so the original test-suite is cross platform based on a ton of locators and environment variables. As long as this will install the version from the build-matrix we're good (the branch version, e.g. 3.13, 3.11)

> Another potential improvement would be to return null from the locator if it can't find a Python installation. This seems to be the expected behavior - see the last line of the LINQ below from PythonEnvironment. However, LocatePythonInternal() throws exceptions when it can't find Python.

Agreed. that should be fixed.

PR #389, opened 2025-04-02
Fix formatting inconsistencies

atifaziz:
This PR addresses several formatting and encoding inconsistencies across the project:

- b099417660d1b03f622a4babd8f5655ca66bd959
  - UTF-8 BOM (`∩╗┐`) markers (removing as that can interfere with some tools) 
- e4a63be208b8d2b4e56de7553acd83c46fbe74d3
  - Mixed CR-LF/LF line encodings that would produce artificially large diffs (as here); 
  - Indentation, spacing, and alignment, etc.
  - Order of imports

I updated `.editorconfig` file to explicitly set the `charset` to `utf-8` for all files. This ensures consistent encoding across the project.  

All problems were then fixed by simply running [`dotnet format`](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-format) (could be interesting to add to CI in the future to avoid these inconsistencies from creeping back).

No functional changes were made to any logic. Files like `src/CSnakes.Runtime.Tests/Converter/BigIntegerConverterTest.cs` where you see a large diff is only due to line endings being normalised.

---
PS Apologies if this seems like minutiae, but it helps to avoid larger than necessary diff changes due to the inconsistencies that often had to be undone locally to keep deltas small.


tonybaloney:
> Apologies if this seems like minutiae
not at all, this has been annoying me for ages.

Do we need to update the git config as well? https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#per-repository-settings

atifaziz:
> Do we need to update the git config as well? https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#per-repository-settings

Good idea and done with 9c3c51a8da617895ec97724b7c9013c56a2629c5, but it's touched about 70% of the files now and the delta says everything in them has changed! We might as well fix this now, but it does mean that open PRs may need to be addressed too as there could be merge conflicts that may be difficult to understand. I can help with that if needed.

PR #393, opened 2025-04-02
Consistent line endings in the repository

gukoff:
Currently, some files are committed with CRLF line endings, and some with LF line endings.

Editing source code on different platforms might automatically change the line endings to suit that platform, adding a lot of noise to git diffs.

This PR changes makes line endings consistent across all platforms. Git will convert text files to LF on commit.

How to do this:

1. Create `.gitattributes` file in the repo root, specify line endings rule
2. Edit existing files with `git add --renormalize .`

See more at: https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings


tonybaloney:
See #389 as well

Issue #377, opened 2025-03-21
MethodReflection.FromMethod needs refactoring


tonybaloney:
It does too much, it's fragile and complicated to the point of being almost unmaintainable.

PR #395, opened 2025-04-03
🚧 Fix Python parameter parsing to reject invalid syntax

atifaziz:
This PR is currently draft and will be published pending:

- [ ] Complete and expand PR description
- [ ] Consider adding legend for short variables or expand them
- [ ] Conduct final polish review (some naming inconsistencies especially with parameter classification)

I'm submitting this draft for visibility to avoid anyone else , as it happened with PR #393 and #389.

---
This PR fixes the Python parameter parsing to reject invalid syntax.

## Benefits

Since more invalid syntax is caught earlier, the source generator can signal those errors as coming from the Python signatures instead of silently generating C# code that may then surface as confusing C# errors in the generated code.

All [tests for invalid parameter syntax from the suite of CPython tests](https://github.com/python/cpython/blob/ffc2f1dd1c023b44b488311511db790a96d757db/Lib/test/test_syntax.py#L385-L537) are now also rejected by CSnake's parser:

| Syntax | Error |
| - | - |
| `None=1` | invalid syntax | 
| `x, y=1, z` | parameter without a default follows parameter with a default | 
| `x, /, y=1, z` | parameter without a default follows parameter with a default | 
| `x, None` | invalid syntax | 
| `*None` | invalid syntax | 
| `**None` | invalid syntax | 
| `/,a,b=,c` | at least one argument must precede / | 
| `a,/,/,b,c` | / may appear only once | 
| `a,/,a1,/,b,c` | / may appear only once | 
| `a=1,/,/,*b,/,c` | / may appear only once | 
| `a,/,a1=1,/,b,c` | / may appear only once | 
| `a,*b,c,/,d,e` | / must be ahead of * | 
| `a=1,*b,c=3,/,d,e` | / must be ahead of * | 
| `a,*b=3,c` | var-positional argument cannot have default value | 
| `a,*b: int=,c` | var-positional argument cannot have default value | 
| `a,**b=3` | var-keyword argument cannot have default value | 
| `a,**b: int=3` | var-keyword argument cannot have default value | 
| `a,*a, b, **c, d` | arguments cannot follow var-keyword argument | 
| `a,*a, b, **c, d=4` | arguments cannot follow var-keyword argument | 
| `a,*a, b, **c, *d` | arguments cannot follow var-keyword argument | 
| `a,*a, b, **c, **d` | arguments cannot follow var-keyword argument | 
| `a=1,/,**b,/,c` | arguments cannot follow var-keyword argument | 
| `*b,*d` | * argument may appear only once | 
| `a,*b,c,*d,*e,c` | * argument may appear only once | 
| `a,b,/,c,*b,c,*d,*e,c` | * argument may appear only once | 
| `a,b,/,c,*b,c,*d,**e` | * argument may appear only once | 
| `a=1,/*,b,c` | expected comma between / and * | 
| `a=1,d=,c` | expected default value expression | 
| `a,d=,c` | expected default value expression | 
| `a,d: int=,c` | expected default value expression | 

It also simplifies the internal management and handling of parameters by the source generator, potentially helping with addressing part of #377.

Types are represented more accurately, like callables, and other superficial ones disappear completely, such as [treating `/` as a special parameter type](https://github.com/tonybaloney/CSnakes/blob/7ca4fb918c11c9a89a777bfe63b7b6b0f84fedc6/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameterType.cs#L7).

## Design

First and foremost, the PR introduces `PythonFunctionParameterList<T>` as way to represent a parameter list with the various parameter types becoming sub-collections and sub-items:

```c#
public sealed class PythonFunctionParameterList<T>(ImmutableArray<T> positional = default,
                                                   ImmutableArray<T> regular = default,
                                                   T? varpos = default,
                                                   ImmutableArray<T> keyword = default,
                                                   T? varkw = default)
{
    public static readonly PythonFunctionParameterList<T> Empty = new();

    public ImmutableArray<T> Positional { get; } = positional.IsDefault ? [] : positional;
    public ImmutableArray<T> Regular { get; } = regular.IsDefault ? [] : regular;
    public T? VariadicPositional { get; } = varpos;
    public ImmutableArray<T> Keyword { get; } = keyword.IsDefault ? [] : keyword;
    public T? VariadicKeyword { get; } = varkw;

    // rest omitted for brevity
}
```

This was directly inspired by Python's own [`ast.arguments`](https://docs.python.org/3/library/ast.html#ast.arguments):

    class ast.arguments(posonlyargs, args, vararg, kwonlyargs, kw_defaults, kwarg, defaults)

`PythonFunctionParameterList<T>` is generic so the actual parameters can be anything, either the nodes produced by parsing or matching `SyntaxNode` required for C# code generation. This means that `FunctionParametersAsParameterSyntaxPairs` goes away:

https://github.com/tonybaloney/CSnakes/blob/7ca4fb918c11c9a89a777bfe63b7b6b0f84fedc6/src/CSnakes.SourceGeneration/Reflection/ArgumentReflection.cs#L101-L114

A (global) type alias, named `PythonFunctionParameterList` is also introduced for `PythonFunctionParameterList<PythonFunctionParameter>` since it's the most common case and makes the code easier on the eyes.

The parser now classifies parameters (positional, regular, keyword-only, args, kwargs) during parse-time. It does this by disambiguating syntax by introduced comma-paired tokens like `, /`, `, *`, `, **`, and `, )`:

```c#
public enum PythonToken
{
    // ...others omitted for previty

    [Token(Example = ", /")]
    CommaSlash,

    [Token(Example = ", *")]
    CommaStar,

    [Token(Example = ", **")]
    CommaStarStar,

    [Token(Example = ", )")]
    CommaCloseParenthesis,
}
```

By doing this, a `,` on its own will be recognised as a separate token than a token composed of adjacent characters like `,/`. This means that parsing of comma-delimited parameters will stop when it encounters `,/` and the parsed parameters can then be classified as positional-only. The same logic happens with `*` and `,*`. As a result, post-processing methods like [`PythonFunctionDefinition.FixupArguments`](https://github.com/tonybaloney/CSnakes/blob/7ca4fb918c11c9a89a777bfe63b7b6b0f84fedc6/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionDefinition.cs#L11-L30) aren't needed anymore. The `,)` token is used in combination with `)` to allow a trailing comma at the end of the parameter list.

As an example of the simplification that this design brings about, see the following code in `MethodReflection`:

```c#
        // Sort the method parameters into this order
        // 1. All positional arguments
        // 2. All keyword-only arguments
        // 3. Any *args argument
        // 4. Any **kwargs argument
        var methodParameters = parameterList
            .Where((a) => a.pythonParameter.ParameterType == PythonFunctionParameterType.Normal && !a.pythonParameter.IsKeywordOnly)
            .Select((a) => a.cSharpParameter)
            .Concat(
                parameterList
                    .Where((a) => a.pythonParameter.ParameterType == PythonFunctionParameterType.Normal && a.pythonParameter.IsKeywordOnly)
                    .Select((a) => a.cSharpParameter)
            ).Concat(
                parameterList
                    .Where((a) => a.pythonParameter.ParameterType == PythonFunctionParameterType.Star)
                    .Select((a) => a.cSharpParameter)
            ).Concat(
                parameterList
                    .Where((a) => a.pythonParameter.ParameterType == PythonFunctionParameterType.DoubleStar)
                    .Select((a) => a.cSharpParameter)
            );
```

which now becomes (partially addressing #377):

```c#
        var methodParameters =
            from ps in new[]
            {
                cSharpParameterList.Positional,
                cSharpParameterList.Regular,
                cSharpParameterList.Keyword,
                cSharpParameterList.VariadicPositional is { } vpd ? [vpd] : [],
                cSharpParameterList.VariadicKeyword is {} vkd ? [vkd] : [],
            }
            from p in ps
            select p;
```

## Summary of Other Changes

- A parameter (`PythonFunctionParameter`) just knows its name and, optionally, type and default value. Members like `IsPositionalOnly` and `IsKeywordOnly` are gone. Where a parameter sits in `PythonFunctionParameterList` decides how it's classified.
- `PythonTypeSpec` changed to be immutable for consistency with other types created during parsing.
- `PythonParser.ConstantValueTokenizer` changed to return a non-nullable `PythonConstant` since its optionality can be handled in the parameter parser.
- Fixed lots of nullability issues
- `QualifiedNamed` no longer allowed as parameter name
- `PythonTypeDefinitionParser` now handles `Callable[..., Any]` correctly


tonybaloney (approved):

tonybaloney:
Fantastic. This is far more robust.

PR #404, opened 2025-04-04
Implement basic class reflection

tonybaloney:
This is stage 1 of #397 

Basic attribute reflection based on names with the ability to override the source name with an attribute.


tonybaloney:
@atifaziz @gukoff  thoughts on this implementation?

atifaziz:
See #584 for an alternative design.

Issue #381, opened 2025-03-27
Support AOT compiled


tonybaloney:
The runtime library violates a few rules for AOT:

```default
3>C:\Users\anthonyshaw\source\repos\tonybaloney\CSnakes\src\CSnakes.Runtime\PyObjectTypeConverter.Generator.cs(23,34,23,89): warning IL3050: Using member 'System.Type.MakeGenericType(params Type[])' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The native code for this instantiation might not be available at runtime.
3>C:\Users\anthonyshaw\source\repos\tonybaloney\CSnakes\src\CSnakes.Runtime\PyObjectTypeConverter.Coroutine.cs(23,34,23,81): warning IL3050: Using member 'System.Type.MakeGenericType(params Type[])' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The native code for this instantiation might not be available at runtime.
3>C:\Users\anthonyshaw\source\repos\tonybaloney\CSnakes\src\CSnakes.Runtime\PyObjectTypeConverter.Dictionary.cs(15,29,15,88): warning IL3050: Using member 'System.Type.MakeGenericType(params Type[])' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The native code for this instantiation might not be available at runtime.
3>C:\Users\anthonyshaw\source\repos\tonybaloney\CSnakes\src\CSnakes.Runtime\PyObjectTypeConverter.List.cs(14,29,14,78): warning IL3050: Using member 'System.Type.MakeGenericType(params Type[])' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The native code for this instantiation might not be available at runtime.
3>C:\Users\anthonyshaw\source\repos\tonybaloney\CSnakes\src\CSnakes.Runtime\PyObjectTypeConverter.Utils.cs(12,30,12,55): warning IL2070: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.Interfaces' in call to 'System.Type.GetInterfaces()'. The parameter 'givenType' of method 'CSnakes.Runtime.PyObjectTypeConverter.IsAssignableToGenericType(Type, Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
[...]

PR #413, opened 2025-04-07
Move known conversions to (AOT-friendly) generated code

atifaziz:
This PR is an initial exploration of how #381 could be addressed, rendering the generated code to work for Native AOT deployments. It does this by directly embedding the conversions at compile-time into the generated code when the types can be known in advance. For the unknown (or presently unhandled) cases, it still relies on delegating to run-time conversion via [`PyObject.As`](https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/CSnakes.Runtime/Python/PyObject.cs#L440-L477).

A side benefit of this approach is that it reduces boxing (and thus heap allocations) for scalars like `int`, `float`, etc.

Pending:

- [x] `int`
- [x] `str`
- [x] `float`
- [x] `bool`
- [x] `bytes`
- [x] `tuple`, `typing.Tuple`, `Tuple`
- [x] `list`, `typing.List`, `List`
- [x] `dict`, `typing.Dict`, `Dict`
- [x] `typing.Mapping`, `Mapping`
- [x] `typing.Sequence`, `Sequence`
- [x] `typing.Coroutine`, `Coroutine` (`async def`)
- [x] `typing.Coroutine`, `Coroutine` (legacy via `@coroutine`)
- [x] `typing.Generator`, `Generator`
- [x] `Buffer`, ~~`typing.Buffer`~~, `collections.abc.Buffer`
- [x] Consider removing local conversion functions and delegate invocation cost
- [x] Rely on caller GIL in `InternalServices.ConvertToCoroutine`
- [ ] Consider emitting warning if `PyObject.As<>` is used for conversion in generated code
- [ ] Choose between parsed and explicit syntax trees for consistency
- [x] Update design notes

### Core Components

- **`IPyObjectImporter<T>`**: This interface defines a contract for importing a Python object into a .NET type `T`. It includes a _static_ abstract method `BareImport(PyObject obj)` that must be implemented by all importers:

  ```c#
  public interface IPyObjectImporter<out T>
  {
      static abstract T BareImport(PyObject obj);
  }
  ```

  This was introduced a while back in PR #204 (with c03cd1be1c1e5b33756d8dcc88ace7f9a3485af6) and so I've simply reused it here except the method was renamed to `BareImport`. The `Bare` prefix is a really there as a reminder that the method is unsafe unless called while the GIL is held.

  Because the method is static, there is no virtual dispatch involved! The cost of the abstraction can be considered as practically zero, so much so that there are instances where compiler can even inline an implementation!

- **`PyObjectImporters` Static Class**: This is the central hub for all importer implementations. It contains nested classes, each implementing `IPyObjectImporter<T>` for specific types or type combinations.

#### Primitive/Scalar Type Importers

These handle basic Python-to-.NET type conversions, such as `bool`, `long`, `double`, `string`, etc. For example, for `bool`, it's as simple as this:

```cs
partial class PyObjectImporters
{
    public sealed class Boolean : IPyObjectImporter<bool>
    {
        public static bool BareImport(PyObject obj) =>
            CPythonAPI.IsPyTrue(obj);
    }
}
```

The above version is simplified for illustration purposes, but each of the actual implementations ensure that the GIL is acquired and throw an `InvalidOperationException` otherwise.

#### Special Importers

`PyObjectImporters.Runtime<T>` is a fallback importer that uses `PyObject.As<T>` to import Python objects of any type. It is useful for cases where the type is not known statically at compile-time or needs dynamic representation and/or execution. Since it _may_ rely on reflection and run-time code generation, it isn't suitable for Native AOT scenarios.

`PyObjectImporters.Clone` simply returns a clone of `PyObject`. It can be seen as a pass-through, or identical to the behavior of `PyObject.As<PyObject>`.

#### Collection Importers

These handle Python collections like lists, dictionaries, and tuples. They use generic parameters to specify the types of elements and their corresponding importers. For example, for lists, it looks like this:

```cs
partial class PyObjectImporters
{
    public sealed class List<T, TImporter> : IPyObjectImporter<IReadOnlyList<T>>
        where TImporter : IPyObjectImporter<T>
    {
        public static IReadOnlyList<T> BareImport(PyObject obj) =>
            new PyList<T, TImporter>(obj.Clone())
    }
}
```

The interesting aspect to note here is that `List` takes _two_ generic type parameters:

1. `T` for the element type
2. `TImporter` for the importer type that knows how to convert a `PyObject` to an element of type `T`; put another way, import a `T` from a `PyObject`

While the second generic type parameter for the importer may seem odd and complex, it enables great flexibility in terms of being able to vary the import strategy for the elements of the list. For example, if `T` is `int`, then `TImporter` can be `PyObjectImporters.Int` or `PyObjectImporters.Runtime<int>`, depending on whether you want to apply a static or run-time strategy. I've taken advantage of this in various places and the same design extends to all other importers for complex types like tuples, dictionaries, coroutines, etc. Hopefully the benefits of this design will become clearer as we go through the examples.

#### Tuple Importers

These are generated dynamically using [T4 templates][t4] to support tuples of varying lengths. Each tuple importer has generic type parameters for each of the tuple's element types as well as their respective importers. Here's how it look for a 2-tuple:

[t4]: https://en.wikipedia.org/wiki/Text_Template_Transformation_Toolkit

```cs
partial class PyObjectImporters
{
    public sealed class Tuple<T1, T2, TImporter1, TImporter2> :
        IPyObjectImporter<(T1, T2)>
        where TImporter1 : IPyObjectImporter<T1>
        where TImporter2 : IPyObjectImporter<T2>
    {
        public static (T1, T2) BareImport(PyObject obj)
        {
            CheckTuple(obj);
            using var a = GetTupleItem(obj, 0);
            using var b = GetTupleItem(obj, 1);
            return (TImporter1.Import(a), TImporter2.Import(b));
        }
    }
}
```

I'll speak more about tuples later as they are a breed apart.

#### Complex Importers

These handle more complex Python objects, like generators and coroutines. For example:

```cs
partial class PyObjectImporters
{
    public sealed class Generator<TYield, TSend, TReturn, TYieldImporter, TReturnImporter> :
        IPyObjectImporter<IGeneratorIterator<TYield, TSend, TReturn>>
        where TYieldImporter : IPyObjectImporter<TYield>
        where TReturnImporter : IPyObjectImporter<TReturn>
    {
        // Implementation omitted for brevity
    }
}
```

Like in the case of `List` shown earlier, not only does the `Generator` importer accept generic type parameters for the _yield_, _send_ and _return_ types, but also for _importer types_ for the yield and return types. This enables mixing and matching import strategies:

```c#
PyObjectImporters.Generator<string, bool, PyObject,
                            PyObjectImporters.String,
                            PyObjectImporters.Clone>
```

for a generator that yields ints and returns a string using the static import strategy.

Note that an importer for `TSend` values isn't there because those are sent into the generator, from .NET to Python, and therefore exported rather than imported.

## Other Design Notes

In the first iteration, the source generator would inline all the code to import each type, which was complicated to maintain and understand in terms of producing syntax trees. Debugging and validation would also be a nightmare as you had inspect the generated code each time. In the second iteration of the design, the source generator emits a sinlge statement only that delegates to an `IPyObjectImporter<T>` implementation for the method return type of `T`. It will emit a single statement even for a complex return type like `tuple[dict[str, list[int]], str, tuple[int, int]]`! The bulk of the import logic (no matter how many composite types are involved) therefore remains in regular types and code in `CSnakes.Runtime`. One can set breakpoints, refactor and debug much more easily. What's more, it's also what prevented from exposing a lot of internal and unsafe functions just for the use in the generated code.

While tuples technically have no limit on the number of elements they can contain, I chose to support tuples with up to 10 elements for this first iteration. This limit should suffice for most practical use cases. For tuples exceeding this limit, the source generator will fall back to runtime importing using `PyObjectImporters.Runtime<T>`. This approach relies on `PyObject.As<T>`,
which is incompatible with Native AOT scenarios. However, this limitation is not inherent to the design and can be addressed in a future PR if needed. For now, it allowed me to reuse the same design instead so there is one strategy to implement and review for all types. The choice of 10 elements was also arbitrary.

## `CSnakes.Runtime` Changes

`CSnakes.Runtime` made several uses of `PyObject.As<>` internally, like here in the constructor for `PythonRuntimeException`:

https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/CSnakes.Runtime/PythonRuntimeException.cs#L10-L20

This means that the code was not compatible with Native AOT scenarios. I've gone ahead and fixed all such cases to use importers instead so that there is no reflection or run-time generic instantiations involved. In fact, the only use of `PyObject.As<>` in all of `CSnakes.Runtime` now appears in `PyObjectImporters.Runtime<T>`.

## Experimantal API

I've taken great care to make sure the new API cannot be reached easily. First of all, it's marked [experimental] so that it will be flagged if used directly. Second, all importers implement `IPyObjectImporter<T>` _explicitly_ so `BareImport` cannot be invoked directly. In fact, you would legally need a _type parameter_!

[experimental]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/experimental-attribute

The generated code still calls `BareImport`, but the experimental warning/error is suppressed. At one point, I even had an additional interface, `PyObject.IUnsafeImportable` to futher shield users from calling `BareImport` without first jumping through hoops, but seeing that this prevented inlining opportunities, I decided to remove it with afe9a9d8da7f8010028811923f45ecfb5ae40a69. Below is the x86-64 disassembly for `TestTwoStrings` that's the proxy for [`test_two_strings` from `test_basic.py`][test_two_strings]. You can see inlining of the `PyUnicode_AsUTF8` from `PyObjectImporters.String` in action:

[test_two_strings]: https://github.com/tonybaloney/CSnakes/blob/07e4e2171d88deac63cf3974e7bef19721e5a915/src/Integration.Tests/python/test_basic.py#L18-L19

```asm
; Assembly listing for method CSnakes.Runtime.TestBasicExtensions+TestBasicInternal:TestTwoStrings(System.String,System.String):System.String:this (FullOpts)
; Emitting BLENDED_CODE for X64 with AVX - Windows
; FullOpts code
; optimized code
; rbp based frame
; fully interruptible
; No PGO data
; 0 inlinees with PGO data; 8 single block inlinees; 0 inlinees without PGO data

G_M000_IG01:                ;; offset=0x0000
       push     rbp
       push     r14
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 112
       lea      rbp, [rsp+0x90]
       xor      eax, eax
       mov      qword ptr [rbp-0x30], rax
       mov      qword ptr [rbp-0x28], rax
       mov      qword ptr [rbp-0x60], rsp
       mov      rbx, rcx
       mov      rsi, rdx
       mov      rdi, r8

G_M000_IG02:                ;; offset=0x0029
       call     [CSnakes.Runtime.Python.GIL:Acquire():System.IDisposable]
       mov      gword ptr [rbp-0x38], rax

G_M000_IG03:                ;; offset=0x0033
       mov      r14, gword ptr [rbx+0x10]
       mov      rcx, 0x7FFDF6FA1E38
       mov      edx, 1
       call     CORINFO_HELP_NEWARR_1_OBJ
       mov      r8, 0x2388010EF50
       mov      gword ptr [rax+0x10], r8
       xor      r8, r8
       mov      gword ptr [rbp-0x30], r8

G_M000_IG04:                ;; offset=0x0060
       mov      dword ptr [rbp-0x28], r8d
       mov      r8, 0x2388010F0E0
       mov      gword ptr [rsp+0x20], r8
       mov      gword ptr [rsp+0x28], rax
       lea      r8, [rbp-0x30]
       mov      rcx, r14
       mov      edx, 1
       xor      r9, r9
       call     [Microsoft.Extensions.Logging.LoggerExtensions:Log(Microsoft.Extensions.Logging.ILogger,int,Microsoft.Extensions.Logging.EventId,System.Exception,System.String,System.Object[])]
       mov      rbx, gword ptr [rbx+0x40]
       mov      rdx, rsi
       mov      rcx, 0x7FFDF74E5278
       call     [CSnakes.Runtime.Python.PyObject:From[System.__Canon](System.__Canon):CSnakes.Runtime.Python.PyObject]
       mov      gword ptr [rbp-0x40], rax

G_M000_IG05:                ;; offset=0x00A8
       mov      rdx, rdi
       mov      rcx, 0x7FFDF74E5278
       call     [CSnakes.Runtime.Python.PyObject:From[System.__Canon](System.__Canon):CSnakes.Runtime.Python.PyObject]
       mov      gword ptr [rbp-0x48], rax

G_M000_IG06:                ;; offset=0x00BF
       mov      rcx, 0x7FFDF74E52D0
       mov      edx, 2
       call     CORINFO_HELP_NEWARR_1_OBJ
       mov      rsi, rax
       lea      rcx, bword ptr [rsi+0x10]
       mov      rdx, gword ptr [rbp-0x40]
       call     CORINFO_HELP_ASSIGN_REF
       lea      rcx, bword ptr [rsi+0x18]
       mov      rdx, gword ptr [rbp-0x48]
       call     CORINFO_HELP_ASSIGN_REF
       cmp      byte  ptr [rbx], bl
       mov      rcx, rbx
       mov      rdx, rsi
       call     [CSnakes.Runtime.Python.PyObject:CallWithArgs(CSnakes.Runtime.Python.PyObject[]):CSnakes.Runtime.Python.PyObject:this]
       mov      gword ptr [rbp-0x50], rax

G_M000_IG07:                ;; offset=0x0102
       cmp      byte  ptr [rax], al
       mov      rcx, rax
       call     [CSnakes.Runtime.CPython.CPythonAPI:PyUnicode_AsUTF8(CSnakes.Runtime.Python.PyObject):System.String]
       mov      rbx, rax

G_M000_IG08:                ;; offset=0x0110
       mov      rcx, gword ptr [rbp-0x50]
       mov      r11, 0x7FFDF6EF0D00
       call     [r11]System.IDisposable:Dispose():this
       nop

G_M000_IG09:                ;; offset=0x0122
       mov      rdx, gword ptr [rbp-0x48]
       test     rdx, rdx
       je       SHORT G_M000_IG11

G_M000_IG10:                ;; offset=0x012B
       mov      rcx, rdx
       mov      r11, 0x7FFDF6EF0D08
       call     [r11]System.IDisposable:Dispose():this
       nop

G_M000_IG11:                ;; offset=0x013C
       mov      rdx, gword ptr [rbp-0x40]
       test     rdx, rdx
       je       SHORT G_M000_IG13

G_M000_IG12:                ;; offset=0x0145
       mov      rcx, rdx
       mov      r11, 0x7FFDF6EF0D10
       call     [r11]System.IDisposable:Dispose():this
       nop

G_M000_IG13:                ;; offset=0x0156
       mov      rax, gword ptr [rbp-0x38]
       test     rax, rax
       je       SHORT G_M000_IG15

G_M000_IG14:                ;; offset=0x015F
       mov      rcx, rax
       mov      r11, 0x7FFDF6EF0D18
       call     [r11]System.IDisposable:Dispose():this

G_M000_IG15:                ;; offset=0x016F
       mov      rax, rbx

G_M000_IG16:                ;; offset=0x0172
       add      rsp, 112
       pop      rbx
       pop      rsi
       pop      rdi
       pop      r14
       pop      rbp
       ret

G_M000_IG17:                ;; offset=0x017D
       push     rbp
       push     r14
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 64
       mov      rbp, qword ptr [rcx+0x30]
       mov      qword ptr [rsp+0x30], rbp
       lea      rbp, [rbp+0x90]

G_M000_IG18:                ;; offset=0x0197
       cmp      gword ptr [rbp-0x50], 0
       je       SHORT G_M000_IG19
       mov      rcx, gword ptr [rbp-0x50]
       mov      r11, 0x7FFDF6EF0D00
       call     [r11]System.IDisposable:Dispose():this

G_M000_IG19:                ;; offset=0x01AF
       nop

G_M000_IG20:                ;; offset=0x01B0
       add      rsp, 64
       pop      rbx
       pop      rsi
       pop      rdi
       pop      r14
       pop      rbp
       ret

G_M000_IG21:                ;; offset=0x01BB
       push     rbp
       push     r14
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 64
       mov      rbp, qword ptr [rcx+0x30]
       mov      qword ptr [rsp+0x30], rbp
       lea      rbp, [rbp+0x90]

G_M000_IG22:                ;; offset=0x01D5
       cmp      gword ptr [rbp-0x48], 0
       je       SHORT G_M000_IG23
       mov      rcx, gword ptr [rbp-0x48]
       mov      r11, 0x7FFDF6EF0D08
       call     [r11]System.IDisposable:Dispose():this

G_M000_IG23:                ;; offset=0x01ED
       nop

G_M000_IG24:                ;; offset=0x01EE
       add      rsp, 64
       pop      rbx
       pop      rsi
       pop      rdi
       pop      r14
       pop      rbp
       ret

G_M000_IG25:                ;; offset=0x01F9
       push     rbp
       push     r14
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 64
       mov      rbp, qword ptr [rcx+0x30]
       mov      qword ptr [rsp+0x30], rbp
       lea      rbp, [rbp+0x90]

G_M000_IG26:                ;; offset=0x0213
       cmp      gword ptr [rbp-0x40], 0
       je       SHORT G_M000_IG27
       mov      rcx, gword ptr [rbp-0x40]
       mov      r11, 0x7FFDF6EF0D10
       call     [r11]System.IDisposable:Dispose():this

G_M000_IG27:                ;; offset=0x022B
       nop

G_M000_IG28:                ;; offset=0x022C
       add      rsp, 64
       pop      rbx
       pop      rsi
       pop      rdi
       pop      r14
       pop      rbp
       ret

G_M000_IG29:                ;; offset=0x0237
       push     rbp
       push     r14
       push     rdi
       push     rsi
       push     rbx
       sub      rsp, 64
       mov      rbp, qword ptr [rcx+0x30]
       mov      qword ptr [rsp+0x30], rbp
       lea      rbp, [rbp+0x90]

G_M000_IG30:                ;; offset=0x0251
       cmp      gword ptr [rbp-0x38], 0
       je       SHORT G_M000_IG31
       mov      rcx, gword ptr [rbp-0x38]
       mov      r11, 0x7FFDF6EF0D18
       call     [r11]System.IDisposable:Dispose():this

G_M000_IG31:                ;; offset=0x0269
       nop

G_M000_IG32:                ;; offset=0x026A
       add      rsp, 64
       pop      rbx
       pop      rsi
       pop      rdi
       pop      r14
       pop      rbp
       ret

; Total bytes of code 629
```

We can leave this API marked as experimental or we can ban their use via an analyzer like [Microsoft.CodeAnalysis.BannedApiAnalyzers]. For now, going with experimental seemed like a simpler starting point.

[Microsoft.CodeAnalysis.BannedApiAnalyzers]: https://www.nuget.org/packages/Microsoft.CodeAnalysis.BannedApiAnalyzers/

<details>

<summary>Initial Design</summary>

## Implementation and Design Notes

⚠️ This is subject to change based on early feedback and as the PR evolves.

It introduces a new abstraction called `IResultConversionCodeGenerator` (design could be revised) that's used by the code generator to emit the conversion code. There are then implementation for each of the known and supported conversions. It also adds conversion helpers for the generated code under `InternalServices`. These are super thin wrappers to existing unsafe and private members of `PyObject` and `CPythonAPI`, but which are necessary to expose. It would be possible to flag their use in user code via an analyzer (along the lines of [Microsoft.CodeAnalysis.BannedApiAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BannedApiAnalyzers/)).

Here's the generated code for something simple like [`test_int_int`](https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/Integration.Tests/python/test_basic.py#L6-L7):

```c#
        public long TestIntInt(long a, long b)
        {
            using (GIL.Acquire())
            {
                logger.LogDebug("Invoking Python function: {FunctionName}", "test_int_int");
                PyObject __underlyingPythonFunc = this.__func_test_int_int;
                using PyObject a_pyObject = PyObject.From(a)!;
                using PyObject b_pyObject = PyObject.From(b)!;
                using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject, b_pyObject);
                var __return = global::CSnakes.Runtime.Python.InternalServices.ConvertToInt64(__result_pyObject);
                return __return;
            }
        }
```

Note the new call to `InternalServices.ConvertToInt64`.

It looks very similar for strings as can be seen in the generated code below for ([`test_two_strings`](https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/Integration.Tests/python/test_basic.py#L18-L19)), except conversion is done by `InternalServices.ConvertToString`:

```c#
        public string TestTwoStrings(string a, string b)
        {
            using (GIL.Acquire())
            {
                logger.LogDebug("Invoking Python function: {FunctionName}", "test_two_strings");
                PyObject __underlyingPythonFunc = this.__func_test_two_strings;
                using PyObject a_pyObject = PyObject.From(a)!;
                using PyObject b_pyObject = PyObject.From(b)!;
                using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject, b_pyObject);
                var __return = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(__result_pyObject);
                return __return;
            }
```

Lists get interesting because they lazily convert items as they are indexed. Below is the generated code for [`test_two_lists_of_strings`](https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/Integration.Tests/python/test_basic.py#L21-L22):

```c#
        public IReadOnlyList<string> TestTwoListsOfStrings(IReadOnlyList<string> a, IReadOnlyList<string> b)
        {
            using (GIL.Acquire())
            {
                logger.LogDebug("Invoking Python function: {FunctionName}", "test_two_lists_of_strings");
                PyObject __underlyingPythonFunc = this.__func_test_two_lists_of_strings;
                using PyObject a_pyObject = PyObject.From(a)!;
                using PyObject b_pyObject = PyObject.From(b)!;
                using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject, b_pyObject);
                static string __tmp_1(PyObject obj)
                {
                    var result = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(obj);
                    return result;
                }

                var __return = global::CSnakes.Runtime.Python.InternalServices.ConvertToList<string>(global::CSnakes.Runtime.Python.InternalServices.Clone(__result_pyObject), __tmp_1);
                return __return;
            }
        }
```

While the list is converted immediately, a local function is created to convert each item and a delegate to it is passed to the list. `PyList<>` has been enhanced to use the delegate if provided or use `PyObject.As` otherwise (the default behaviour for run-time cases).

Finally, the code generated for tuples, as show in the case of [`tuple_3`](https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/Integration.Tests/python/test_tuples.py#L7-L8) below, demonstrates full nesting of conversion code in action: 

```c#
        public (string, string, string) Tuple3((string, string, string) a)
        {
            using (GIL.Acquire())
            {
                logger.LogDebug("Invoking Python function: {FunctionName}", "tuple_3");
                PyObject __underlyingPythonFunc = this.__func_tuple_3;
                using PyObject a_pyObject = PyObject.From(a)!;
                using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject);
                global::CSnakes.Runtime.Python.InternalServices.CheckTuple(__result_pyObject);
                using var __tmp_1 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 0);
                var __tmp_2 = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(__tmp_1);
                using var __tmp_3 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 1);
                var __tmp_4 = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(__tmp_3);
                using var __tmp_5 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 2);
                var __tmp_6 = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(__tmp_5);
                var __return = (__tmp_2, __tmp_4, __tmp_6);
                return __return;
            }
        }
```

</details>


tonybaloney:
Really interesting, I like the idea. I tried refactoring `As<>` a few times but couldn't get past nested types without a generic.

At a glance I see the following pros:

- Faster code (having analysed the benchmarks)
- Clearer boxing

Possible cons:

- Exposing unsafe methods which have unexpected behaviours. 
- Exposing methods which require the GIL (everything beneath `.As<>`

Keen to see what something like `tuple[dict[str, list[int]], str, tuple[int, int]]` looks like!

latop2604:
Look like this and works

```csharp
public (IReadOnlyDictionary<string, IReadOnlyList<long>>, string, (long, long)) Big((IReadOnlyDictionary<string, IReadOnlyList<long>>, string, (long, long)) a)
{
    using (GIL.Acquire())
    {
        logger.LogDebug("Invoking Python function: {FunctionName}", "big");
        PyObject __underlyingPythonFunc = this.__func_big;
        using PyObject a_pyObject = PyObject.From(a)!;
        using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject);
        global::CSnakes.Runtime.Python.InternalServices.CheckTuple(__result_pyObject);
        using var __tmp_1 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 0);
        IReadOnlyDictionary<string, IReadOnlyList<long>> __tmp_2 = __tmp_1.As<IReadOnlyDictionary<string, IReadOnlyList<long>>>();
        using var __tmp_3 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 1);
        var __tmp_4 = global::CSnakes.Runtime.Python.InternalServices.ConvertToString(__tmp_3);
        using var __tmp_5 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__result_pyObject, 2);
        global::CSnakes.Runtime.Python.InternalServices.CheckTuple(__tmp_5);
        using var __tmp_7 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__tmp_5, 0);
        var __tmp_8 = global::CSnakes.Runtime.Python.InternalServices.ConvertToInt64(__tmp_7);
        using var __tmp_9 = global::CSnakes.Runtime.Python.InternalServices.GetTupleItem(__tmp_5, 1);
        var __tmp_10 = global::CSnakes.Runtime.Python.InternalServices.ConvertToInt64(__tmp_9);
        var __tmp_6 = (__tmp_8, __tmp_10);
        var __return = (__tmp_2, __tmp_4, __tmp_6);
        return __return;
    }
}
```

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
```diff
@@ -0,0 +1,153 @@
...
+public static partial class PyObjectImporters
+{
+    public sealed class Runtime<T> : IPyObjectImporter<T>
+    {
+        public static T Import(PyObject obj) => obj.As<T>();
+    }
+
+    public sealed class Boolean : IPyObjectImporter<bool>
+    {
+        public static bool Import(PyObject obj) => CPythonAPI.IsPyTrue(obj);
```
tonybaloney, line 29: This requires the GIL to be held

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public sealed class Int64 : IPyObjectImporter<long>
+    {
+        public static long Import(PyObject obj) => CPythonAPI.PyLong_AsLongLong(obj);
```
tonybaloney, line 34: This requires the GIL to be held

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public sealed class Double : IPyObjectImporter<double>
+    {
+        public static double Import(PyObject obj) => CPythonAPI.PyFloat_AsDouble(obj);
```
tonybaloney, line 39: This requires the GIL to be held

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public sealed class Single : IPyObjectImporter<float>
+    {
+        public static float Import(PyObject obj) => (float)CPythonAPI.PyFloat_AsDouble(obj);
```
tonybaloney, line 44: This requires the GIL to be held

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    public sealed class String : IPyObjectImporter<string>
+    {
+        public static string Import(PyObject obj) => CPythonAPI.PyUnicode_AsUTF8(obj);
```
tonybaloney, line 49: This requires the GIL to be held

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
AaronRobinsonMSFT: Is there an assert that can be applied to check? All the uses I've seen below are holding the lock.

src/CSnakes.SourceGeneration/ResultConversionCodeGenerator.cs
```diff
@@ -3,197 +3,129 @@
...
-                                                GenericName(Identifier("As"), TypeArgumentList(SingletonSeparatedList(TypeSyntax))))))))))
-        ];
+        switch (pythonTypeSpec)
+        {
+            case { Name: "int" }: return Long;
+            case { Name: "str" }: return String;
+            case { Name: "float" }: return Double;
+            case { Name: "bool" }: return Boolean;
+            case { Name: "bytes" }: return ByteArray;
+            case { Name: "Buffer" or "typing.Buffer" or "collections.abc.Buffer" }: return Buffer;
```
AaronRobinsonMSFT, line 43: Where are these sorts of things bubbling up from?

src/CSnakes.SourceGeneration/ResultConversionCodeGenerator.cs
```diff
@@ -3,197 +3,129 @@
...
+            }
+            case { Name: "typing.Sequence" or "Sequence", Arguments: [var t] }:
+            {
+                return ListConversionGenerator(t, "Sequence");
+            }
+            case { Name: "tuple" or "typing.Tuple" or "Tuple", Arguments: { Length: >= 1 and <= 12 } ts }:
+            {
+                var generators = ImmutableArray.CreateRange(from t in ts select Create(t));
+                return new ConversionGenerator(TupleType(SeparatedList(from item in generators select TupleElement(item.TypeSyntax))),
+                                               TypeReflection.CreateGenericType("Tuple", [.. from item in generators select item.TypeSyntax, .. from item in generators select item.ImporterTypeSyntax]));
```
AaronRobinsonMSFT, line 57: I defer to @tonybaloney for the final say, but this mixture of linq and pattern matching is very complicated to tease out exactly the intent or if there is a logical bug.

AaronRobinsonMSFT: I guess my biggest gripe is this is not debuggable. So if it is correct, w00t. If not, good luck stepping through it.

src/CSnakes.SourceGeneration/ResultConversionCodeGenerator.cs
AaronRobinsonMSFT: I speaking specifically about the `typing.` and `collections.abc.` prefixes.

atifaziz: [`Buffer` lives in the `collections.abc`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer) module. The match for the naked `Buffer` is if someone has done `from collections.abc import Buffer` earlier in the Python module (though it's an assumption since the parser doesn't validate because imports imply code execution). `typing.Buffer` looks wrong to be honest. [Some aliases used to live in `typing`, but got moved](https://docs.python.org/3/library/typing.html#deprecated-aliases), but this doesn't look like one of them. It always lived in `collections.abc`, right @tonybaloney? To be honest, I took this pattern directly from `TypeReflection.AsPredefinedType` without questioning it for maintaining maximum behavioural compatibility in this PR:

https://github.com/tonybaloney/CSnakes/blob/26febb004e1543d2c3bc2d71da11451a0d41c67e/src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs#L16-L33

Apart from `typing.Buffer`, all other `typing` prefixes are there to support legacy aliases. That said, I do think they can be removed since CSnakes only supports Python 3.9+ and the earliest deprecations happened in that version. At the same time, the deprecated versions are harmless can also be removed later.

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
atifaziz: > This requires the GIL to be held, as do all of the `CPythonAPI` calls in this class. Calling any of these methods without the GIL will likely cause a SEGFAULT or similar

I'm aware of this, but all this publicly visible code is housed under the `Internals` namespace that is considered as reserved for CSnake's own use and currently all the import code invocation is called under the GIL in the generated code.

I hadn't marked this PR as fully ready for review yet but happy to get some early feedback (in fact one of the reasons why I opened it up in draft). I was planning to wrap it up this weekend, post the design notes and also call out things (such as this) as open for debate.

atifaziz:
@tonybaloney I'm taking this PR out of draft. The good news is that the new design simpler, cleaner, maintainable and more optimised (hope you'll agree). Here's a quick rundown of all the improvements and changes:

- Only a single line is added to the generated code
- No more generation of local functions or delegates
- Zero-cost abstraction for importing Python objects; no virtual dispatch
- Made it fairly hard (though not impossible) to invoke so-called unsafe functions (defined as those needing GIL though they fail in safe ways)
- `InternalServices` class is completely gone
- New (small surface) API is:
  - marked experimental so will be flagged if used directly
  - strong-typed so most incorrect usages will be caught at compile-time
- No special handling for coroutines; in fact, the generated code for .NET does not even need use of `async` and `await` so not it simplifies `MethodReflection` (adding to #377)
- Some internal code paths in `CSnakes.Runtime` could have lead to prevent Native AOT scenarios, but these are now addressed as well by using the new design internally

src/CSnakes.Runtime/Python/Internals/PyObjectImporters.cs
atifaziz: This was asserted with 9874b8b513671edcd350057a27da149d797cb5e4, but is now even checked in release builds with afe9a9d8da7f8010028811923f45ecfb5ae40a69.

AaronRobinsonMSFT:
@atifaziz This looks good. There are still issues in src/CSnakes.Runtime/PyObjectTypeConverter.Generator.cs around the `MakeGenericType()`. Thoughts on mitigating that in a manner so if the fallback path is taken, the warning fires and isn't triggered in the regular case?

atifaziz:
> There are still issues in src/CSnakes.Runtime/PyObjectTypeConverter.Generator.cs around the `MakeGenericType()`. Thoughts on mitigating that in a manner so if the fallback path is taken, the warning fires and isn't triggered in the regular case?

@AaronRobinsonMSFT This PR was about rendering the source-generated code friendly for Native AOT deployments. If the Python code is well type-annotated and the C# proxies strong-typed, then it would be disappointing for anyone to learn that their code breaks at run-time for Native AOT. For issue you've raised, I think we build on top of what's being added here. I've opened PR #416 as draft, where you can see a prototype of how we could deal with and completely circumvent `PyObject.As<T>` through [interceptors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#interceptors) during compilation source generation. We could argue then whether `PyObject.As<T>()` is worth the trouble and whether it should be removed in favour of encouraging something more static like `PyObject.ImportAs<T, TImporter>` (helpers could be added for common types to make it more palatable). That then leaves `PyObject.As(Type)`, but I imagine that would be designed for super dynamic scenarios and should be marked with `[RequiresDynamicCodeAttribute]` so as to alert users making accidental use. That said, one code still use interception to accommodate Native AOT scenarios there if the type is known at compile-time, as in `PyObject.As(typeof(string))`. I'm just not sure it'd be worth the trouble for those (supposedly) rare instances.

src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_basic.approved.txt
```diff
@@ -134,7 +136,8 @@ public static class TestClassExtensions
                 using PyObject a_pyObject = PyObject.From(a)!;
                 using PyObject b_pyObject = PyObject.From(b)!;
                 using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject, b_pyObject);
-                return __result_pyObject.As<double>();
+                var __return = __result_pyObject.BareImportAs<double, global::CSnakes.Runtime.Python.PyObjectImporters.Double>();
```
tonybaloney, line 139: Why does it need the full namespace?

tonybaloney:
Can you rebase this too, it's missing the `IntConverterTests`

tonybaloney:
Before

| Method                    | Mean        | Error       | StdDev      |
|-------------------------- |------------:|------------:|------------:|
| ComplexReturn             |  4,497.4 ns |    87.83 ns |   151.51 ns |
| ComplexReturnLazy         |  3,793.3 ns |    73.84 ns |    90.68 ns |
| FunctionReturnsList       |  4,056.1 ns |    79.37 ns |   125.89 ns |
| FunctionTakesList         | 32,412.2 ns |   511.24 ns |   478.22 ns |
| FunctionTakesDictionary   | 72,389.4 ns | 1,375.05 ns | 1,286.22 ns |
| FunctionReturnsDictionary | 16,062.7 ns |   161.43 ns |   151.01 ns |
| FunctionReturnsTuple      |  1,189.4 ns |     4.30 ns |     3.81 ns |
| FunctionTakesTuple        |  2,039.7 ns |    39.97 ns |    44.43 ns |
| FunctionTakesValueTypes   |    777.0 ns |     1.51 ns |     1.41 ns |
| EmptyFunction             |    233.9 ns |     0.55 ns |     0.51 ns |

After

| Method                    | Mean        | Error     | StdDev    |
|-------------------------- |------------:|----------:|----------:|
| ComplexReturn             |  3,087.1 ns |  28.34 ns |  25.12 ns |
| ComplexReturnLazy         |  2,898.0 ns |  44.34 ns |  41.47 ns |
| FunctionReturnsList       |  3,702.7 ns |  73.82 ns |  95.99 ns |
| FunctionTakesList         | 26,664.4 ns | 449.00 ns | 419.99 ns |
| FunctionTakesDictionary   | 50,932.4 ns | 168.03 ns | 157.18 ns |
| FunctionReturnsDictionary | 15,958.7 ns |  72.30 ns |  64.09 ns |
| FunctionReturnsTuple      |    783.7 ns |   1.70 ns |   1.59 ns |
| FunctionTakesTuple        |  1,468.9 ns |   7.57 ns |   7.08 ns |
| FunctionTakesValueTypes   |    773.8 ns |   4.62 ns |   4.32 ns |
| EmptyFunction             |    231.4 ns |   0.43 ns |   0.40 ns |

tonybaloney:
`FunctionReturnsTuple` is the benchmark I'd expect an improvement in (and is showing a good speedup), I'm not sure why `FunctionTakesList` and `FunctionTakesDictionary` would be faster?

src/CSnakes.Runtime.Tests/Converter/ListConverterTest.cs
```diff
@@ -1,8 +1,15 @@
...
+                      PyObjectImporters.List<long, PyObjectImporters.Int64>,
+                      ListConverterTest>,
+    IConverterTestCasesContainer<IReadOnlyList<long>>
 {
-    [Fact]
-    public void ListConverter() =>
-        RunTest<IReadOnlyList<long>>([123456, 123562]);
+    public static TheoryData<IReadOnlyList<long>> TestCases => new()
+    {
+        new[] { 123456L, 123562 }
```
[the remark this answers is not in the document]
atifaziz: It wasn't unintentional either. 😅 C# infers `long` based on any one of the elements being long. If it's confusing, happy to add the `L` suffix to the second as well or specify the element type explicitly (as in, `new long[] { 123456, 123562 }`).

src/CSnakes.Runtime.Tests/Converter/TupleConverterTests.cs
```diff
@@ -1,40 +1,52 @@
...
-        (long, long, long) input = (42, 42, 42);
-        RunTest(input);
-    }
+public class Tuple3ConverterTests :
+    ConverterTestBase<(long, long, long),
+        PyObjectImporters.Tuple<long, long, long, PyObjectImporters.Int64, PyObjectImporters.Int64, PyObjectImporters.Int64>,
+        Tuple3ConverterTests>,
+    IConverterTestCasesContainer<(long, long, long)>
+{
+    public static TheoryData<(long, long, long)> TestCases => new() { (42, 42, 42) };
```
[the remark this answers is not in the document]
atifaziz: Usually I try to avoid making unrelated changes in the same PR so I maintained the data here as it is in `main` for now:

https://github.com/tonybaloney/CSnakes/blob/052c614392501981f5f8bfd7c1ee7025f0ddc14c/src/CSnakes.Runtime.Tests/Converter/TupleConverterTests.cs#L19-L24

src/CSnakes.Runtime.Tests/Converter/TupleConverterTests.cs
```diff
@@ -1,40 +1,52 @@
...
-        (long, long) input = (42, 42);
-        RunTest(input);
-    }
+public class Tuple2ConverterTests :
+    ConverterTestBase<(long, long),
+        PyObjectImporters.Tuple<long, long, PyObjectImporters.Int64, PyObjectImporters.Int64>,
+        Tuple2ConverterTests>,
+    IConverterTestCasesContainer<(long, long)>
+{
+    public static TheoryData<(long, long)> TestCases => new() { (42, 42) };
```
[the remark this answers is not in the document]
atifaziz: Usually I try to avoid making unrelated changes in the same PR so I maintained the data here as it is in `main` for now:

https://github.com/tonybaloney/CSnakes/blob/052c614392501981f5f8bfd7c1ee7025f0ddc14c/src/CSnakes.Runtime.Tests/Converter/TupleConverterTests.cs#L13-L17

src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_basic.approved.txt
atifaziz: It doesn't, but good practice to avoid potential conflicts with the user project (although this is very rare).

atifaziz:
> I'm not sure why `FunctionTakesList` and `FunctionTakesDictionary` would be faster?

The list and dictionary elements get lazily initialised so that can't be it. My guess therefore would be that the cost of `MakeGenericType` is the bulk that goes away. That's especially true when any one of the type arguments to `MakeGenericType` is a value type since those would need a new concrete implementation to be generated and JIT-ed.

atifaziz:
> Can you rebase this too, it's missing the `IntConverterTests`

Done with 14c8862df670c68adfecd278dbe62a82c22a9628.

tonybaloney (approved):

tonybaloney:
@AaronRobinsonMSFT any final comments before I merge this?

AaronRobinsonMSFT:
Sorry @tonybaloney. Nope, this looks good. Thanks @atifaziz

src/CSnakes.Runtime.Tests/Converter/TupleConverterTests.cs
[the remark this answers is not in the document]
atifaziz: Follow-up in PR #428.

PR #415, opened 2025-04-13
Fix `PyObject.As` for `int` on LP64 & LLP64

atifaziz:
`PyObject.As` for `int`/`Int32` ends up calling [`PyLong_AsLong`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong), which returns a `long`, but a `long` has different widths on different 64-bit platforms. For example, Windows uses the LLP64 data model where a `long` is 32 bits wide and Linux uses LP64 where a `long` is 64 bits wide. This PR fixes the problem by using [`PyLong_AsLongLong`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong) instead since it returns `long long` and that's consistently 64 bits wide across LLP64 and LP64 data models. On the .NET side, `long long` pops up as `Int64`, which is then converted to `int`/`Int32` with an overflow check. This change makes the behaviour consistent.

This fix also indirectly addresses another (more surprising) problem. When an `int` is sent to Python and then converted back, you get an `InvalidCastException` with the message:

> Unable to cast object of type 'System.Int64' to type 'System.Int32'.

### DIFF
--- a/src/CSnakes.Runtime.Tests/Converter/IntConverterTest.cs
+++ b/src/CSnakes.Runtime.Tests/Converter/IntConverterTest.cs
@@ -19,7 +19,8 @@ public class IntConverterTest : ConverterTestBase
     [InlineData(int.MinValue - 1L)]
     public void TestOverflow(long input)
     {
-        void Act() => _ = PyObject.From(input).As<int>();
+        using var n = PyObject.From(input);
+        void Act() => _ = n.As<int>();
         _ = Assert.Throws<OverflowException>(Act);
     }
 }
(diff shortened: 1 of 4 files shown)


tonybaloney:
Oh. I thought I fixed this ages ago. From memory, the behaviour between macOS and Linux isn't the same. 
The Buffer extension methods have a switch in there for the bit width.

tonybaloney:
Ignore me. You're talking about `int32`

CPython have a discussion thread on using C STD types like uint32_t that would make everyone's lives easier but for now we have this weird API. 

Python `int` back to C# `int` is more likely to cause overflow errors, which we have tests for. But I don't expect most C# devs to know that an int is equivalent to a BigInteger

PR #417, opened 2025-04-16
Return immortals from `PyObject.From` for `long` + `BigInteger`

atifaziz:
`PyObject.From` would return immortals for 0, 1 and -1 as long as they were `int` values. It would allocate a new `PyObject` if 0, 1 and -1 were typed as `long` or `BigInteger`. This PR extends `PyObject.From` to also consider `long` and `BigInteger` values. Tests have been extended for both cases (refactored to break out each case and consolidate all duplication). At the same time, all immortals ard considered for returning outside of taking the GIL, which I believe is safe. `PyObject.From` takes the GIL when it's exhausted those optimisations or immortal cases.

---
PS I believe some of the tests are failing on Linux ARM and macOS due to the problem that #415 fixes.

### DIFF
--- a/src/CSnakes.Runtime.Tests/Python/ImmortalTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/ImmortalTests.cs
@@ -1,60 +1,93 @@
 using CSnakes.Runtime.Python;
 
 namespace CSnakes.Runtime.Tests.Python;
-public class ImmortalTests : RuntimeTestBase
+
+public interface IImmortalFromTestCasesContainer<T>
 {
+    static abstract TheoryData<T> TestCases { get; }
+}
+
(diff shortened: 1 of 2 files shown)


tonybaloney:
Build errors suggest I stuffed up that merge commit? https://github.com/tonybaloney/CSnakes/pull/417/commits/016333354074c782d6c60c795f3c11bc14e7598d

atifaziz:
> Build errors suggest I stuffed up that merge commit? [0163333](https://github.com/tonybaloney/CSnakes/commit/016333354074c782d6c60c795f3c11bc14e7598d)

Fixed with db814fdcb361baf41ecc5108b92eef993160b007. All checks passing.

tonybaloney (approved):

Issue #353, opened 2025-02-16
Expose cancellation tokens for generated async methods


tonybaloney:
The underlying Coroutine object accepts a cancellation token, but the source generator doesn't expose the cancellation token in the generated methods.

PR #424, opened 2025-04-23
Fix `CancellationToken` parameters to be non-nullable

atifaziz:
This PR fixes `CancellationToken` parameters in member related to coroutines to be non-nullable, which is more standard. The documentation for [`CancellationToken.None`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.none?view=net-9.0) has more on this:

> The cancellation token returned by this property cannot be canceled; that is, its [`CanBeCanceled`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.canbecanceled?view=net-9.0) property is `false`.
>
> You can also use the C# [`default(CancellationToken)`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/default) statement to create an empty cancellation token.
>
> Two empty cancellation tokens are always equal.

Making a `CancellationToken` nullable unnecessarily adds a third state, which is redundant with an empty one.

### DIFF
--- a/docs/async_support.md
+++ b/docs/async_support.md
@@ -49,4 +49,4 @@ See [Free-Threading Mode](advanced.md#free-threading-mode) for more information.
 
 ## Cancellation Tokens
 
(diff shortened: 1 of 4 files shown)


tonybaloney:
Related #353

tonybaloney (approved):

PR #427, opened 2025-04-23
Skip C# proxies for underscored functions

atifaziz:
This PR skips code generation of C# proxies for Python functions whose name starts with an underscore (`_`). As per [PEP 8](https://peps.python.org/pep-0008/), these are conventionally considered _internal_:

> - `_single_leading_underscore`: weak “internal use” indicator. E.g. `from M import *` does not import objects whose names start with an underscore.


tonybaloney (approved):
Good idea

PR #430, opened 2025-04-25
Make GIL acquisition a `ref struct`

atifaziz:
This PR proposes to return the GIL acquisition as a `ref struct` called ~~`AcquiredLock`~~ `GIL.Lock`. The semantics are as before except the `ref struct` ensures that the lock cannot escape to the heap easily when state machines get formed indirectly in code using `async`/`await` and `yield`. Either way, it prevents the code from compiling if the code after a `yield` or `await` could resume on a different thread than on which the lock was acquired.

The following code was possible up to now:

```c#
static IEnumerable<int> Test()
{
    using (GIL.Acquire())
        yield return 42;
}

static async Task<int> TestAsync()
{
    using (GIL.Acquire())
        await Task.Yield();
    return 42;
}
```

but will produce error CS4007 after this PR:

    error CS4007: Instance of type 'CSnakes.Runtime.Python.GIL.Lock' cannot be preserved across 'await' or 'yield' boundary.

The fix would be to do as follows instead:

```c#
static IEnumerable<int> Test()
{
    using (GIL.Acquire())
    {
        // ...do something...
    }

    yield return 42;

    using (GIL.Acquire())
    {
        // ...do something more...
    }
}

static async Task<int> TestAsync()
{
    using (GIL.Acquire())
    {
        // ...do something...
    }

    await Task.Yield();

    using (GIL.Acquire())
    {
        // ...do something more...
        return 42;
    }

}
```

### DIFF
--- a/src/CSnakes.Runtime/Python/GIL.cs
+++ b/src/CSnakes.Runtime/Python/GIL.cs
@@ -83,18 +83,18 @@ public static class GIL
         public int RecursionCount => recursionCount;
     }
 
-    public readonly ref struct AcquiredLock : IDisposable
+    public readonly ref struct Lock : IDisposable
     {
         private readonly PyGilState state;
 
-        internal AcquiredLock(PyGilState state) => this.state = state;
+        internal Lock(PyGilState state) => this.state = state;
 
         public void Dispose() => state.Dispose();
 
-        internal bool Equals(AcquiredLock other) => state == other.state;
+        internal bool Equals(Lock other) => state == other.state;
     }
 
-    public static AcquiredLock Acquire()
+    public static Lock Acquire()
     {
         if (currentState == null)
         {
--- a/src/CSnakes.Runtime/Python/GIL.cs
+++ b/src/CSnakes.Runtime/Python/GIL.cs
@@ -83,7 +83,20 @@ public static class GIL
         public int RecursionCount => recursionCount;
     }
 
-    public static IDisposable Acquire()
+#pragma warning disable IDE0250 // Make struct 'readonly' (logically read-write)
+    public ref struct AcquiredLock : IDisposable
+#pragma warning restore IDE0250 // Make struct 'readonly'
+    {
+        private readonly PyGilState state;
+
+        internal AcquiredLock(PyGilState state) => this.state = state;
+
+        public void Dispose() => state.Dispose();
+
+        internal bool Equals(AcquiredLock other) => state == other.state;
+    }
+
+    public static AcquiredLock Acquire()
     {
         if (currentState == null)
         {
@@ -93,7 +106,7 @@ public static class GIL
         {
             currentState.Ensure();
         }
-        return currentState;
+        return new(currentState);
     }
 
     internal static void QueueForDisposal(nint handle)
--- a/src/CSnakes.Runtime/Python/GIL.cs
+++ b/src/CSnakes.Runtime/Python/GIL.cs
@@ -83,9 +83,7 @@ public static class GIL
         public int RecursionCount => recursionCount;
     }
 
-#pragma warning disable IDE0250 // Make struct 'readonly' (logically read-write)
-    public ref struct AcquiredLock : IDisposable
-#pragma warning restore IDE0250 // Make struct 'readonly'
+    public readonly ref struct AcquiredLock : IDisposable
     {
         private readonly PyGilState state;
 
--- a/src/CSnakes.Runtime.Tests/Python/GILTests.cs
+++ b/src/CSnakes.Runtime.Tests/Python/GILTests.cs
@@ -14,7 +14,6 @@ public class GILTests : RuntimeTestBase
     {
         using var gil = GIL.Acquire();
 
-        Assert.NotNull(gil);
         Assert.True(GIL.IsAcquired);
     }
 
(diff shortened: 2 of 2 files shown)


tonybaloney:
This is outside of my expertise. Tagging Aaron R for review

src/CSnakes.Runtime/Python/GIL.cs
```diff
@@ -83,7 +83,20 @@ public void Dispose()
         public int RecursionCount => recursionCount;
     }
 
-    public static IDisposable Acquire()
+#pragma warning disable IDE0250 // Make struct 'readonly' (logically read-write)
```
AaronRobinsonMSFT, line 86: Not trying to start a big debate here, but this type isn't read-write. A field of it is being modified but not the field itself, so the type should be marked `readonly`. I'm not sure if the C# language team has chimed in on this and likely should, but for now I would marked this as `readonly` also.

Note that if `PyGilState` was a value type, I would agree with the above warning suppresion.

AaronRobinsonMSFT:
@tonybaloney So semantically this is a desirable change. This creates a compiler error when .NET magic can cause a thread transition - which if I recall is not good for the GIL lock. That is a win. A downside to this model is it is now very easy to make mistakes by missing to take the GIL lock, so worse UX. Note that this is correct based on the constraints of the GIL and its thread affinity semantics.

An alternative approach to consider is embrace the rules of Python and supply your own thread scheduler that doesn't allow for switching during `async`/`await`. Unclear how this would impact existing performance of users.

If this is entirely within the CSnakes implementation and _not_ public, it makes perfect sense. Making it public might cause some friction.

atifaziz:
Thanks for your comments and thoughts, @AaronRobinsonMSFT.

> when .NET magic can cause a thread transition - which if I recall is not good for the GIL lock.

Definitely not and can lead to fatal errors like this:

    Fatal Python error: PyEval_SaveThread: the function must be called with the GIL held, after Python initialization and before Python finalization, but the GIL is released (the current Python thread state is NULL)
    Python runtime state: initialized
    
    Thread 0x00006618 (most recent call first):
      <no Python frame>
    
    Thread 0x0000c928 (most recent call first):
      <no Python frame>

> A downside to this model is it is now very easy to make mistakes by missing to take the GIL lock, so worse UX.

But this isn't specific to _this_ proposal or design, right? Missing to take the GIL would be fatal irrespective.

> An alternative approach to consider is embrace the rules of Python and supply your own thread scheduler that doesn't allow for switching during `async`/`await`. Unclear how this would impact existing performance of users.

While that could make sense for `async`/`await`, but I wouldn't go so far for `yield`.

> If this is entirely within the CSnakes implementation and _not_ public, it makes perfect sense. Making it public might cause some friction.

I think the only reason `GIL` is public is because it's used in the generated code otherwise it's not expected to be consumed from user code. Consequently, this proposal is really to help add guardrails for CSnakes maintainers and contributors to prevent a class of bugs.

See the implementation of the `PyEnumerable<,>` iterator method in PR #432 for how clumsy and hairy the GIL acquisition and disposal looks (not to mention to reason about) to release and re-acquire across `yield` boundaries:

https://github.com/tonybaloney/CSnakes/blob/4844e30e4604e154f49b0717b5cdb762441318cf/src/CSnakes.Runtime/Python/PyEnumerable.cs#L15-L50

With this PR, the implementation will look simple to read, reason about and review:

```c#
private static IEnumerator<TValue> Iterator(PyObject pyIterator)
{
    using var _ = pyIterator;

    while (true)
    {
        TValue import;

        using (GIL.Acquire())
        {
            nint result = CPythonAPI.PyIter_Next(pyIterator);
            if (result == IntPtr.Zero)
            {
                if (CPythonAPI.PyErr_Occurred())
                    throw PyObject.ThrowPythonExceptionAsClrException();

                yield break;
            }

            using (var itemObject = PyObject.Create(result))
                import = TImporter.BareImport(itemObject);
        }

        yield return import;
    }
}
```

Any incorrect use will cause the code to be rejected by the compiler due to error CS4007.

src/CSnakes.Runtime/Python/GIL.cs
atifaziz: > I'm not sure if the C# language team has chimed in on this and likely should, but for now I would marked this as `readonly` also.

I think the way they've chimed in on this is that `readonly` doesn't mean _pure_:

> It is worth noting that _readonly instance member_ != _pure instance member_. A _pure_ instance member guarantees no state will be modified. A `readonly` instance member only guarantees that instance state will not be modified.

So fair call and added the `readonly` modifier with 8c6a4bb82da571f2cf7c23bdf7e90438898a8c3d.

AaronRobinsonMSFT:
> But this isn't specific to this proposal or design, right? Missing to take the GIL would be fatal irrespective.

Not really, but if we could handle the thread pool logic then a simple single `using` is much easier to reason about than the proposed model. The approach here is absolutely correct, no doubt. I'm simply pointing out that the thread pool switching is the root cause here and that could technically be "fixed".

> I think the only reason GIL is public is because it's used in the generated code otherwise it's not expected to be consumed from user code.

Right, so technically `public`, but practically implementation detail. Once `UnsafeAccessorType` supports `ref struct`s we could fix this. Sadly, no sooner than .NET 12 I think.

AaronRobinsonMSFT (approved):
LGTM

PR #437, opened 2025-04-29
Implement truthy-falsy for `PyObject`

atifaziz:
## Overview

This PR adds support for Python's truthy-falsy behavior to the `PyObject` class in CSnakes. This enhancement allows C# code to use Python objects directly in boolean contexts (conditionals, logical operations) in a way that mimics Python's behavior.

## Background

In Python, objects can be evaluated in boolean contexts, following specific rules for truthiness:

- Objects like empty collections (`[]`, `{}`, `()`), zero (`0`), `None`, and empty strings (`""`) evaluate to `False`
- Non-empty collections, non-zero numbers, non-empty strings, and most other objects evaluate to `True`

This PR implements this behavior for CSnakes, allowing seamless integration of Python's boolean semantics in C#.

## Technical Details

The implementation adds:

1. Three operator overloads in `PyObject`:
   - `operator true` - Determines if a Python object is truthy
   - `operator !` - Implements logical negation, determining if an object is falsy
   - `operator false` - Required by C# when `operator true` is defined, but simply delegates to `operator !`

2. Corresponding native function bindings to CPython's API:
   - `PyObject_IsTrue` - Returns 1 if the object is considered true, 0 if false
   - `PyObject_Not` - Returns 1 if the object is considered false, 0 if true

3. Comprehensive test cases covering various Python types and their expected boolean evaluation:
   - Basic values (`True`, `False`, `None`)
   - Numeric values (negative, zero, positive)
   - Strings (empty, non-empty)
   - Collections (empty and non-empty lists, dictionaries, tuples)

**Note on `operator false`**: C# requires both `operator true` and `operator false` to be defined together when overloading Boolean operators. However, `operator false` is not directly testable, but it is somewhat indirectly covered by tests for `operator !` since `operator false` just delegates to `operator !`.

## Usage Example

```csharp
using var pyList = PyObject.From(new[] { 1, 2, 3 });

// Python-style boolean evaluation
if (pyList)
{
    // Will execute because non-empty lists are truthy
}

using var pyEmptyDict = PyObject.From(new Dictionary<string, object>());

// Python-style logical not
if (!pyEmptyDict)
{
    // Will execute because empty dictionaries are falsy
}
```

This enhancement improves the ergonomics of working with Python objects in C#, making the API more natural and Pythonic.

### DIFF
--- a/src/CSnakes.Runtime/Python/PyObject.cs
+++ b/src/CSnakes.Runtime/Python/PyObject.cs
@@ -304,6 +304,43 @@ public partial class PyObject : SafeHandle, ICloneable
         }
     }
 
+    /// <summary>
+    /// Check if the object is false. Equivalent to the <c>not</c> operator in Python.
+    /// </summary>
+    public static bool operator !(PyObject obj)
+    {
+        using (GIL.Acquire())
+        {
+            return CPythonAPI.PyObject_Not(obj) switch
+            {
+                < 0 => throw ThrowPythonExceptionAsClrException(),
+                0 => false,
+                _ => true,
+            };
+        }
+    }
+
+    /// <summary>
+    /// Check if the object is true. Equivalent to the <c>bool()</c> function in Python.
+    /// </summary>
+    public static bool operator true(PyObject obj)
+    {
+        using (GIL.Acquire())
+        {
+            return CPythonAPI.PyObject_IsTrue(obj) switch
+            {
+                < 0 => throw ThrowPythonExceptionAsClrException(),
+                0 => false,
+                _ => true,
+            };
+        }
+    }
+
+    /// <summary>
+    /// Check if the object is false. Equivalent to the <c>not</c> operator in Python.
+    /// </summary>
+    public static bool operator false(PyObject obj) => !obj;
+
     public override int GetHashCode()
     {
         using (GIL.Acquire())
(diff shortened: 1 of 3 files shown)


src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -304,6 +304,49 @@ private static bool Compare(PyObject left, PyObject right, CPythonAPI.RichCompar
...
     }
 
+    /// <summary>
+    /// Check if the object is false. Equivalent to the <c>not</c> operator in Python.
+    /// </summary>
+    public static bool operator !(PyObject obj)
+    {
+        if (obj.Is(False))
+            return true;
+
```
tonybaloney, line 314: Add a shortcut condition for None as well

tonybaloney (changes requested):

src/CSnakes.Runtime/Python/PyObject.cs
atifaziz: Good point and done with 0ccbfe30e57797ca6a84c27a52418b11e71aca30.

src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -304,6 +304,49 @@ private static bool Compare(PyObject left, PyObject right, CPythonAPI.RichCompar
...
+    }
+
+    /// <summary>
+    /// Check if the object is true. Equivalent to the <c>bool()</c> function in Python.
+    /// </summary>
+    public static bool operator true(PyObject obj)
+    {
+        if (obj.Is(True))
+            return true;
+
```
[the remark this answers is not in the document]
atifaziz: Good point and done with 0ccbfe30e57797ca6a84c27a52418b11e71aca30.

tonybaloney (approved):

PR #438, opened 2025-04-29
Refactor to use a single event loop for coroutines

atifaziz:
## Summary

This PR refactors the Python coroutine handling in CSnakes to use a single shared event loop _and_ thread for all coroutines instead of creating a new one for each coroutine operation and per thread. The change significantly improves performance and operational behaviour (more on that later), especially when running multiple coroutines concurrently.

## Design Overview

The refactored implementation:

- uses a single, lazily-initialized, shared event loop that runs continuously.
- implements a message queue pattern for scheduling coroutines on the shared loop.
- runs the event loop on a dedicated background thread via `Task.Run`.
- supports cancellation, failure and successful completion of tasks.

### Key Components

1. **Shared Event Loop**
   - A singleton `EventLoop` instance is created on demand and maintained until disposed with finalisation
   - Uses a simple lock-based lazy initialization pattern through `GetDefaultEventLoop()`
   - The [event loop runs forever](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_forever) until it is [stopped/interrupted](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.stop) by some change that's signaled via a message arriving on the queue.

2. **Message Queue**
   - A `ConcurrentQueue<Request>` is used for communication with the event loop thread.
   - Scheduling of new coroutines and cancellation of existing tasks can happen on any thread, so [`call_soon_threadsafe`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_soon_threadsafe) is used to safely move these operations to the thread servicing the event loop.
   - Different messages are sent to the queue to signal:
     - scheduling of a new task
     - completion of task, whether it got cancelled, failed or completed
     - stopping of the event loop (when getting disposed)

3. **Task Management**
   - `CoroutineTask` class wraps Python tasks and handles their lifecycle
   - Support for cancellation and exception handling
   - Proper disposal and resource cleanup

4. **Compatible API**
   - The public `Coroutine<TYield, TSend, TReturn>.AsTask(CancellationToken)` API remains unchanged
   - Internal implementation is significantly simplified and more robust

## Performance Improvements

At the micro-level, benchmarks demonstrate significant performance improvements, especially as number concurrent operations increase:

| Coroutine Count | Performance Improvement | Memory Overhead |
|-----------------|-------------------------|-----------------|
| 1               | 10% faster              | 764B → 1.45KB |
| 10              | 61% faster              | 6.6KB → 13.8KB |
| 100             | 69% faster              | 61KB → 134KB |
| 1000            | 76% faster              | 601KB → 1.3MB |

This is despite making more calls into Python and the _O(N<sup>2</sup>) time complexity of scanning and disposing done tasks.

While memory usage doubles, the performance improvement is substantial, especially at scale. This trade-off is well worth it for any application running multiple Python coroutines, which won't be uncommon. One must bear in mind that there is also less native memory pressure due to only a single thread and event loop being used for all coroutines. So on the balance of things, the effective and working set memory usage may not be exactly double.

The detailed benchmark is as follows:

| Diff    | Method          | N    |                 Mean | Error          |      Allocated |
| ------- | --------------- | ---- | -------------------: | -------------- | -------------: |
| Old     | `AsyncFunction` | 1    |             15.51 ms | 0.045 ms       |          764 B |
| **New** |                 |      | **14.035 ms (-10%)** | **0.4669 ms**  |    **1.45 KB** |
| Old     | `AsyncFunction` | 10   |             15.57 ms | 0.035 ms       |         6643 B |
| **New** |                 |      |  **5.998 ms (-61%)** | **0.8086 ms**  |   **13.78 KB** |
| Old     | `AsyncFunction` | 100  |             62.51 ms | 1.243 ms       |        61243 B |
| **New** |                 |      | **19.470 ms (-69%)** | **2.8330 ms**  |  **134.16 KB** |
| Old     | `AsyncFunction` | 1000 |            295.31 ms | 18.179 ms      |       601752 B |
| **New** |                 |      | **72.220 ms (-76%)** | **14.1494 ms** | **1335.27 KB** |

At the macro-level, the improvements are even more significant. Using a single even loop yields more correct and predictable operation. With the current design, increasing the sleep delay in `AsyncFunction` to 1 second and then launching 1,000 coroutines concurrently takes 47 seconds and over 20 threads (and an event loop per thread) from the thread pool to complete. With the refactored design, all thousand coroutines complete in 1.07 seconds, which is reasonable to expect! If there was a timeout associated with the overall operation of say 10 seconds, the former would always time-out and therefore never complete!

## Ideas for future/further improvements

- Adding event loop statistics and monitoring capabilities
- More efficient tracking of the task that just got done, reducing the _O(N<sup>2</sup>) time complexity
- Reduce memory usage for very large coroutine counts

### DIFF
--- a/src/CSnakes.Runtime/CPython/EventLoop.cs
+++ b/src/CSnakes.Runtime/CPython/EventLoop.cs
@@ -1,0 +1,322 @@
+using CSnakes.Runtime.Python;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+
+namespace CSnakes.Runtime.CPython;
+internal sealed class EventLoop : IDisposable
+{
+    private bool disposed;
+    private readonly PyObject loop = CPythonAPI.NewEventLoop();
+    private Methods methods;
+    private Task runForeverTask;
+    private readonly ConcurrentQueue<Request> requestQueue = new();
+    private readonly PyObject taskLoopStopFunction;
+
+    private abstract class Request : IDisposable
+    {
+        public virtual void Dispose() { }
+    }
+
+    private sealed class ScheduleRequest(PyObject coroutine, CancellationToken cancellationToken) : Request
+    {
+        public PyObject Coroutine { get; } = coroutine.Clone();
+        public CancellationToken CancellationToken { get; } = cancellationToken;
+        public TaskCompletionSource<TaskCompletionSource<PyObject>> CompletionSource { get; } = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        public override void Dispose() => coroutine.Dispose();
+    }
+
+    private sealed class CancelRequest(CoroutineTask coroTask, CancellationToken cancellationToken) : Request
+    {
+        public CoroutineTask Task { get; } = coroTask;
+        public CancellationToken CancellationToken { get; } = cancellationToken;
+    }
+
+    private sealed class StopRequest : Request
+    {
+        public static readonly StopRequest Instance = new();
+
+        private StopRequest() { }
+    }
+
+    private sealed class CoroutineTask(PyObject pyTask) : IDisposable
+    {
+        private readonly PyObject pyTask = pyTask;
+        private PyObject? doneMethod;
+        private CancellationToken cancellationToken;
+
+        public TaskCompletionSource<PyObject> CompletionSource { get; } = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        /// <remarks>
+        /// Since <c>done</c> is expected to be called many times, cache the the
+        /// method object to avoid the overhead of calling <see
+        /// cref="PyObject.GetAttr"/> each time.
+        /// </remarks>
+        private PyObject DoneMethod => this.doneMethod ??= pyTask.GetAttr("done");
+
+        public void Dispose()
+        {
+            this.doneMethod?.Dispose();
+            this.pyTask.Dispose();
+        }
+
+        public void Cancel(CancellationToken cancellationToken = default)
+        {
+            using var cancelMethod = pyTask.GetAttr("cancel");
+            cancelMethod.Call().Dispose();
+            this.cancellationToken = cancellationToken;
+        }
+
+        public bool Conclude()
+        {
+            if (ProcessConclusion() == TaskStatus.Running)
+                return false;
+
+            Dispose();
+            return true;
+        }
+
+        private TaskStatus ProcessConclusion()
+        {
+            if (this.pyTask.IsClosed)
+                return CompletionSource.Task.Status;
+
+            // If the task is not done yet, return without doing anything.
+
+            if (DoneMethod.Call().Is(PyObject.False)) // TODO Use falsy-ness when PR #437 is merged
+                return TaskStatus.Running;
+
+            // If the Python task is cancelled, set the corresponding .NET task to cancelled.
+
+            using (var cancelledMethod = this.pyTask.GetAttr("cancelled"))
+            {
+                if (cancelledMethod.Call().Is(PyObject.True)) // TODO Use truthy-ness when PR #437 is merged
+                {
+                    CompletionSource.SetCanceled(this.cancellationToken);
+                    return TaskStatus.Canceled;
+                }
+            }
+
+            // If the Python task raised an exception, set the corresponding .NET task to faulted.
+
+            using (var exceptionMethod = this.pyTask.GetAttr("exception"))
+            {
+                PyObject? exception = exceptionMethod.Call();
+                try
+                {
+                    if (!exception.IsNone())
+                    {
+                        using var type = exception.GetPythonType();
+                        string name = type.GetAttr("__name__").ImportAs<string, PyObjectImporters.String>();
+                        CompletionSource.SetException(new PythonInvocationException(name, exception, null));
+                        exception = null;
+                        return TaskStatus.Faulted;
+                    }
+                }
+                finally
+                {
+                    exception?.Dispose();
+                }
+            }
+
+            // If the Python task is finished, set the result in the corresponding .NET task.
+
+            using var resultFunction = this.pyTask.GetAttr("result");
+            PyObject? result = resultFunction.Call();
+            try
+            {
+                CompletionSource.SetResult(result);
+                result = null; // yield ownership
+                return TaskStatus.RanToCompletion;
+            }
+            finally
+            {
+                result?.Dispose();
+            }
+        }
+    }
+
+    public static EventLoop RunNewForever() => new();
+
+    private EventLoop()
+    {
+        this.methods = new Methods(this.loop);
+        using var vars = PyObject.Create(CPythonAPI.PyDict_New());
+        this.taskLoopStopFunction = CPythonAPI.PyRun_String("lambda task: task.get_loop().stop()", CPythonAPI.InputType.Py_eval_input, vars, vars);
+        this.runForeverTask = Task.Run(RunForever);
+    }
+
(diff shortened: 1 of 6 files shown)


atifaziz:
Benchmarks from 02ff30273b87e1a34a41b3aca75030ceaa4deb73:

| Method          | N    | Delay |         Mean |      Error |     StdDev |    Gen0 |  Allocated |
| --------------- | ---- | ----- | -----------: | ---------: | ---------: | ------: | ---------: |
| `AsyncFunction` | 1    | 0.001 |    15.333 ms |  0.3052 ms |  0.4074 ms |       - |    1.56 KB |
| `AsyncFunction` | 1    | 1     | 1,008.631 ms |  1.7731 ms |  1.6586 ms |       - |    2.22 KB |
| `AsyncFunction` | 10   | 0.001 |     8.097 ms |  1.0320 ms |  3.0429 ms |       - |   14.82 KB |
| `AsyncFunction` | 10   | 1     | 1,009.712 ms |  1.7997 ms |  1.5954 ms |       - |   15.88 KB |
| `AsyncFunction` | 100  | 0.001 |    11.273 ms |  1.9347 ms |  5.5821 ms |       - |  147.45 KB |
| `AsyncFunction` | 100  | 1     | 1,006.105 ms |  6.0190 ms |  5.0261 ms |       - |  144.58 KB |
| `AsyncFunction` | 1000 | 0.001 |    47.898 ms |  0.9555 ms |  2.4147 ms | 90.9091 |  1468.6 KB |
| `AsyncFunction` | 1000 | 1     | 1,039.493 ms | 12.9844 ms | 11.5104 ms |       - | 3703.76 KB |

---

- BenchmarkDotNet v0.14.0
- Windows 11 (10.0.26100.3775)
- 13th Gen Intel Core i7-1370P, 1 CPU, 20 logical and 14 physical cores
- .NET SDK 9.0.203
  - **Host**: .NET 9.0.4 (9.0.425.16305), X64 RyuJIT AVX2
  - DefaultJob: .NET 9.0.4 (9.0.425.16305), X64 RyuJIT AVX2

atifaziz:
Benchmarks from 02ff30273b87e1a34a41b3aca75030ceaa4deb73:

| Method          | N    | Delay |         Mean |      Error |     StdDev |    Gen0 |  Allocated |
| --------------- | ---- | ----- | -----------: | ---------: | ---------: | ------: | ---------: |
| `AsyncFunction` | 1    | 0.001 |    15.333 ms |  0.3052 ms |  0.4074 ms |       - |    1.56 KB |
| `AsyncFunction` | 1    | 1     | 1,008.631 ms |  1.7731 ms |  1.6586 ms |       - |    2.22 KB |
| `AsyncFunction` | 10   | 0.001 |     8.097 ms |  1.0320 ms |  3.0429 ms |       - |   14.82 KB |
| `AsyncFunction` | 10   | 1     | 1,009.712 ms |  1.7997 ms |  1.5954 ms |       - |   15.88 KB |
| `AsyncFunction` | 100  | 0.001 |    11.273 ms |  1.9347 ms |  5.5821 ms |       - |  147.45 KB |
| `AsyncFunction` | 100  | 1     | 1,006.105 ms |  6.0190 ms |  5.0261 ms |       - |  144.58 KB |
| `AsyncFunction` | 1000 | 0.001 |    47.898 ms |  0.9555 ms |  2.4147 ms | 90.9091 |  1468.6 KB |
| `AsyncFunction` | 1000 | 1     | 1,039.493 ms | 12.9844 ms | 11.5104 ms |       - | 3703.76 KB |

---

- BenchmarkDotNet v0.14.0
- Windows 11 (10.0.26100.3775)
- 13th Gen Intel Core i7-1370P, 1 CPU, 20 logical and 14 physical cores
- .NET SDK 9.0.203
  - **Host**: .NET 9.0.4 (9.0.425.16305), X64 RyuJIT AVX2
  - DefaultJob: .NET 9.0.4 (9.0.425.16305), X64 RyuJIT AVX2

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        private TaskStatus ProcessConclusion()
+        {
+            if (this.pyTask.IsClosed)
+                return CompletionSource.Task.Status;
+
+            // If the task is not done yet, return without doing anything.
+
+            if (DoneMethod.Call().Is(PyObject.False)) // TODO Use falsy-ness when PR #437 is merged
```
tonybaloney, line 86:
```suggestion
            if (!DoneMethod.Call())
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -0,0 +1,322 @@
...
+                return TaskStatus.Running;
+
+            // If the Python task is cancelled, set the corresponding .NET task to cancelled.
+
+            using (var cancelledMethod = this.pyTask.GetAttr("cancelled"))
+            {
+                if (cancelledMethod.Call().Is(PyObject.True)) // TODO Use truthy-ness when PR #437 is merged
```
tonybaloney, line 93:
```suggestion
                if (cancelledMethod.Call())
```

tonybaloney (changes requested):

tonybaloney (approved):

src/CSnakes.Runtime/CPython/EventLoop.cs
atifaziz: Done with 0dbc4086b451ace76283f80efe3eb4d1901736e1.

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -0,0 +1,322 @@
...
+            using (var exceptionMethod = this.pyTask.GetAttr("exception"))
+            {
+                PyObject? exception = exceptionMethod.Call();
+                try
+                {
+                    if (!exception.IsNone())
+                    {
+                        using var type = exception.GetPythonType();
+                        string name = type.GetAttr("__name__").ImportAs<string, PyObjectImporters.String>();
+                        CompletionSource.SetException(new PythonInvocationException(name, exception, null));
```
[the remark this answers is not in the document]
atifaziz: Noted in 0dbc4086b451ace76283f80efe3eb4d1901736e1.

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -0,0 +1,322 @@
...
+
+    public void Dispose()
+    {
+        if (this.disposed)
+            return;
+
+        try
+        {
+            Enqueue(StopRequest.Instance);
+            this.runForeverTask.WaitAsync(TimeSpan.FromSeconds(10)).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
atifaziz: An arbitrary decision to be honest. This is happening inside `Dispose`, which is synchronous (although we should also consider implementing `IAsyncDisposable` but that doesn't change the problem) and the shutdown should be fast (cancelling and draining/disposing tasks). If it takes as long as 10 seconds, even under heavy loads, then it most probably points to a freeze and a bug. Since we don't want to hang the host due to an internal implementation error, the timeout or any other error is logged (in debug builds only) and the method proceeds to dispose the objects before returning to the caller. The tricky thing here is that `Dispose` should not throw exceptions and yet we have some non-trivial work to do here if there are tasks in flight, so we do have to swallow and move on somehow. This is something that could be given more thought. If you have ideas on how to handle this better, then I'm all ears.

PR #442, opened 2025-04-30
async iterators

tonybaloney:


atifaziz:
@tonybaloney Looks like `PyObject_GetAIter` was only added in 3.10 so all test runs using 3.9 are failing. May need to call `__aiter__` directly in that case.

PR #443, opened 2025-05-03
Remove redundant GIL acquisition during `PyObject` release

atifaziz:
This PR removes a redundant GIL acquisition during the release of a `PyObject`.

### DIFF
--- a/src/CSnakes.Runtime/Python/PyObject.cs
+++ b/src/CSnakes.Runtime/Python/PyObject.cs
@@ -45,10 +45,7 @@ public partial class PyObject : SafeHandle, ICloneable
         }
         if (GIL.IsAcquired)
         {
-            using (GIL.Acquire())
-            {
-                CPythonAPI.Py_DecRefRaw(handle);
-            }
+            CPythonAPI.Py_DecRefRaw(handle);
         }
         else
         {


tonybaloney:
The GIL is required and I want to avoid a possible race condition from the check above. 

The scenario is a custom `__del__` method on a heap allocated class which would be called when the object is disposed and the GIL assertion would cause a hard crash. It's possible we're just not testing that right now (or aren't testing the race)

atifaziz:
> The GIL is required and I want to avoid a possible race condition from the check above.
> 
> The scenario is a custom `__del__` method on a heap allocated class which would be called when the object is disposed and the GIL assertion would cause a hard crash. It's possible we're just not testing that right now (or aren't testing the race)

If it's this tricky, then I think we should add a comment, but I am still having a tough time seeing the race condition. The `GIL` class is designed with TLS in mind for avoiding nested acquisitions for the same thread. If `GIL.IsAcquired` returned `ture`, it means that the calling thread already has the GIL. If there's a thread-switch between the `GIL.IsAcquired` check and `GIL.Acquire` then it wouldn't matter because other threads would see their own state of GIL _acquisition_. When the interrupted thread resumes, the result of its `GIL.IsAcquired` won't have changed. Am I missing something?

PR #446, opened 2025-05-04
Make all loggers & logging optional

atifaziz:
This PR makes all loggers & logging optional:

- All `ILogger` instances across the codebase are made nullable.
- Logging calls are updated to handle `null` gracefully via the null-conditional operator (`?.`)
- Logging services are no longer required ([`GetRequiredService`] calls replaced with [`GetService`] where the requested service is a logger)

There are 3 reasons for this change:

- Logging should be optional
- Application should not break (via `GetRequiredService`) if logging isn't available as a service
- Logging should not incur any cost if no logger is given or available.

[`GetRequiredService`]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getrequiredservice?view=net-9.0-pp
[`GetService`]: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getservice?view=net-9.0-pp

### DIFF
--- a/src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
+++ b/src/CSnakes.Runtime/EnvironmentManagement/CondaEnvironmentManagement.cs
@@ -4,10 +4,10 @@ using System.Runtime.InteropServices;
 
 namespace CSnakes.Runtime.EnvironmentManagement;
 #pragma warning disable CS9113 // Parameter is unread. There for future use.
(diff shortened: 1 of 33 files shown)


tonybaloney (approved):

tonybaloney:
Seems like a good idea and would help toward the requirement to work without the Microsoft IoC Container

atifaziz:
> would help toward the requirement to work without the Microsoft IoC Container

You read my mind.

Issue #450, opened 2025-05-06
Embed module sources into generated interfaces


tonybaloney:
Distributing packages is tricky when you need to give the path to `PYTHONHOME` which includes the user code.

We should support an (optional) toggle to embed the source text of the Python code that was used during the generation process.

tonybaloney:
Depends on #457

PR #451, opened 2025-05-06
Optionally embed source code for modules in generated classes

tonybaloney:
Implements #450 

Also adds a public API for importing a module from a string.

Although not all sources will be UTF-8, it uses base64 to encode the string into UTF-8 byte array, then base64 encodes it and decodes back into UTF-8. UTF-8 is the standard in Python. UTF-16 and UTF-16-LE are common in .NET projects (especially Windows) but fiddly to get working in Python.


src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -122,7 +139,14 @@ from f in functionNames
                         using (GIL.Acquire())
                         {
                             logger?.LogDebug("Importing module {ModuleName}", "{{fileName}}");
-                            module = Import.ImportModule("{{fileName}}");
+                            if (!string.IsNullOrEmpty(encodedSource))
+                            {
+                                module = Import.ImportModule("{{fileName}}", Encoding.UTF8.GetString(Convert.FromBase64String(encodedSource)), "{{fileName}}.py");
```
atifaziz, line 144: What's the purpose of encoding the Python source into Base64? I'd be strongly inclined to embed the Python source in clear text, as-is.

src/RedistributablePython.Tests/RedistributablePython.Tests.csproj
atifaziz, line 1: I think this change was accidental, so undo?

src/RedistributablePython.Tests/.globalconfig
atifaziz, line 1: What effect is this supposed to have?

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -92,6 +107,8 @@ public static class {{pascalFileName}}Extensions
 
                 private static ReadOnlySpan<byte> HotReloadHash => "{{HexString(hash.AsSpan())}}"u8;
 
+                private const string? encodedSource = "{{base64Value}}";
```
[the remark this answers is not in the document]
atifaziz: nit: I'd move this `const` to the bottom, so C# source comes before the embedded Python source otherwise one would have to potentially scroll through a lot to get to the desired bits (assuming someone navigated to this generated file to inspect the C# code).

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -122,7 +139,14 @@ from f in functionNames
...
                             logger?.LogDebug("Importing module {ModuleName}", "{{fileName}}");
-                            module = Import.ImportModule("{{fileName}}");
+                            if (!string.IsNullOrEmpty(encodedSource))
+                            {
+                                module = Import.ImportModule("{{fileName}}", Encoding.UTF8.GetString(Convert.FromBase64String(encodedSource)), "{{fileName}}.py");
+                            }
+                            else
+                            {
+                                module = Import.ImportModule("{{fileName}}");
+                            }
```
atifaziz, line 149: nit: Consider a conditional expression:

```suggestion
                            this.module = !string.IsNullOrEmpty(source)
                                        ? Import.ImportModule("{{fileName}}", Encoding.UTF8.GetString(Convert.FromBase64String(encodedSource)), "{{fileName}}.py");
                                        : Import.ImportModule("{{fileName}}");
```

src/CSnakes.Runtime/CPython/Import.cs
```diff
@@ -56,4 +80,6 @@ protected static nint GetBuiltin(string name)
     [LibraryImport(PythonLibraryName)]
     internal static partial nint PyImport_ReloadModule(PyObject module);
 
+    [LibraryImport(PythonLibraryName, StringMarshalling = StringMarshalling.Custom, StringMarshallingCustomType = typeof(NonFreeUtf8StringMarshaller))]
```
atifaziz, line 83: I don't think `NonFreeUtf8StringMarshaller` is the correct choice here because it would _leak_ the strings! I think `NonFreeUtf8StringMarshaller` was designed for exceptional cases like `Py_GetVersion` that returns a pointer to static storage and [`PyUnicode_AsUTF8`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8), whose return value isn't caller-owned:

> This caches the UTF-8 representation of the string in the Unicode object, and subsequent calls will return a pointer to the same buffer. The caller is not responsible for deallocating the buffer. The buffer is deallocated and pointers to it become invalid when the Unicode object is garbage collected.

[`Py_CompileString`](https://docs.python.org/3/c-api/veryhigh.html#c.Py_CompileString) doesn't mention anything about string ownership so I think it's safe to assume they're caller owns them and responsible for freeing (i.e. the marshalling layer in .NET).

src/CSnakes.Runtime/CPython/Import.cs
```diff
@@ -18,6 +18,27 @@ internal static PyObject Import(string name)
...
+        
+        nint codeObject = Py_CompileString(code, path, InputType.Py_file_input);
+        if (codeObject == IntPtr.Zero)
+        {
+            throw PyObject.ThrowPythonExceptionAsClrException();
+        }
+
+        nint pyName = AsPyUnicodeObject(name);
+        nint pyCode = AsPyUnicodeObject(code);
+        nint pyPath = AsPyUnicodeObject(path);
```
atifaziz, line 32: Why not use [`PyObject.From(string)`](https://github.com/tonybaloney/CSnakes/blob/35db15a98c5a4b604b900fabb9325909277b103e/src/CSnakes.Runtime/Python/PyObject.cs#L598-L607) for these and where the `Py_DecRefRaw` calls below could also be handled through `using` declarations and disposal?

src/CSnakes.Runtime/CPython/Import.cs
```diff
@@ -18,6 +18,27 @@ internal static PyObject Import(string name)
...
     }
 
+    internal static PyObject Import(string name, string code, string path)
+    {
+        
+        nint codeObject = Py_CompileString(code, path, InputType.Py_file_input);
+        if (codeObject == IntPtr.Zero)
+        {
+            throw PyObject.ThrowPythonExceptionAsClrException();
+        }
```
atifaziz, line 28: This error case logic is already handled by [`PyObject`'s constructor and `Create`](https://github.com/tonybaloney/CSnakes/blob/35db15a98c5a4b604b900fabb9325909277b103e/src/CSnakes.Runtime/Python/PyObject.cs#L16-L28), so why not just do `PyObject.Create` on the returned value?

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
tonybaloney: To preserve the encoding and not have issues with the encoding of the source file interfering with the original source. Also it takes up less space.

src/RedistributablePython.Tests/.globalconfig
tonybaloney: This enables the feature for the Redistributable Tests

src/RedistributablePython.Tests/RedistributablePython.Tests.csproj
tonybaloney: The path should be relative

src/CSnakes.Runtime/CPython/Import.cs
tonybaloney: Lets come back to this.

atifaziz: Okay, but then let's track it in an issue so it's not forgotten?

tonybaloney: #452

Issue #452, opened 2025-05-07
(DEV) Review usage of non free UTF8 marshaller


tonybaloney:
We have a custom string marshalling class for `string` -> `const char*`

Review where it is used and if it is being used correctly.

atifaziz:
Something to investigate, for example:

https://github.com/tonybaloney/CSnakes/pull/451#discussion_r2076115380

atifaziz:
This can be closed on ce #455 and #458 are merged.

PR #455, opened 2025-05-07
Reduce allocations, leaks & decoding when embedding source

atifaziz:
This PR brings many benefits:

- Stores the Python source code directly as a [UTF-8 string literal](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals) so it's visible in the generated code.
- C# will optimise the storage by storing the source code bytes directly in the binary's `.data` section instead of the [string intern pool](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-string-intern).
-  The loading and usage will be allocation-free and copy-free as the span will directly point into the binary where the data is stored and that pointer is subsequently passed to Python via [`ReadOnlySpanMarshaller`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalling.readonlyspanmarshaller-2?view=net-9.0).
- The run-time cost of decoding Base64 and encoding UTF-8 disappears.
- The potential memory leak due to use of `NonFreeUtf8StringMarshaller` is gone (partially addressing #452).
- There is no empty constant or conditional guard against it in the generated code.

A new test has been added to validate the handling of various Unicode code points and flag corruption due to encoding issues; See:

- `src/Integration.Tests/python/test_source.py`
- `src/Integration.Tests/SourceTests.cs`


src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_source.approved.txt
```diff
@@ -0,0 +1,131 @@
...
+    /// def get_mad_string() -> str: ...
+    /// ]]></code>
+    /// </summary>
+    string GetMadString();
+}
+
+file static class ThisModule
+{
+    private static ReadOnlySpan<byte> source => """"
+        def get_mad_string() -> str:
```
tonybaloney, line 113: So how does it handle the indentation of the source file?

src/Integration.Tests/python/test_source.py
```diff
@@ -0,0 +1,13 @@
+def get_mad_string() -> str:
```
tonybaloney, line 1: Include a doc string (triple quote) in here for good measure.

src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_source.approved.txt
atifaziz: From [Raw string literal text - `"""` in string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string):

> The following rules govern the interpretation of a multi-line raw string literal:
> 
> - The opening quotes must be the last non-comment token on its respective line, and the closing quote must be the first non-comment token on its respective line.
> - **Any whitespace to the left of the closing quotes is removed from all lines of the raw string literal.**
> - Whitespace following the opening quote on the same line is ignored.
> - Whitespace only lines following the opening quote are included in the string literal.
> - If a whitespace precedes the end delimiter on the same line, the exact number and kind of whitespace characters (e.g. spaces vs. tabs) must exist at the beginning of each content line. Specifically, a space does not match a horizontal tab, and vice versa.
> - The newline before the closing quotes isn't included in the literal string.

src/Integration.Tests/python/test_source.py
atifaziz: Done with 8f9cd7d3b90865fd5991623790a0f3d61caed730.

tonybaloney (approved):

PR #470, opened 2025-05-24
Change all spawned processes to use argument lists. Add `InstallPackage` to `IPythonPackageInstaller`

### LINKED ISSUE #465: Error creating Virtual Environment if path contains spaces
**Actual Result**
- _.WithVirtualEnvironment(path)_ does not work is path contains spaces (e.g.: C:\Users\user\AppData\Local\My Framework\My application\Python\\.venv

**Expected Result**
- _.WithVirtualEnvironment(path)_ should support path contains spaces

**Additional Information**
- _.WithVirtualEnvironment(path)_ works fine If path is C:\Users\user\AppData\Local\MyFramework\Myapplication\Python\\.venv

tonybaloney:
This changes the ProcessUtils class to always use `ArgumentList` instead of `Arguments`.

This will reduce the risk of applications being vulnerable to shell injection, and it will also fix paths with spaces.

Fixes #465

This change also introduces another method to `IPythonPackageInstaller` called `InstallPackage` for installing a single package given a name.

### DIFF
--- a/src/CSnakes.Runtime/ProcessUtils.cs
+++ b/src/CSnakes.Runtime/ProcessUtils.cs
@@ -10,13 +10,9 @@ internal static class ProcessUtils
     internal static (Process proc, string? result, string? errors) ExecutePythonCommand(ILogger? logger, PythonLocationMetadata pythonLocation, params string[] arguments)
     {
 
-        ProcessStartInfo startInfo = new()
+        ProcessStartInfo startInfo = new(pythonLocation.PythonBinaryPath, arguments)
         {
             WorkingDirectory = pythonLocation.Folder,
-            FileName = pythonLocation.PythonBinaryPath,
-            ArgumentList = {
-                // TODO
-            },
             RedirectStandardError = true,
             RedirectStandardOutput = true,
             CreateNoWindow = true,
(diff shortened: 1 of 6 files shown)


src/CSnakes.Runtime/ProcessUtils.cs
```diff
@@ -1,45 +1,41 @@
...
 using Microsoft.Extensions.Logging;
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 
 namespace CSnakes.Runtime;
 
 internal static class ProcessUtils
 {
-    internal static (Process proc, string? result, string? errors) ExecutePythonCommand(ILogger? logger, PythonLocationMetadata pythonLocation, string arguments)
+    internal static (Process proc, string? result, string? errors) ExecutePythonCommand(ILogger? logger, PythonLocationMetadata pythonLocation, params string[] arguments)
```
atifaziz, line 10: Consider `ReadOnlySpan<string>` for the variadic `arguments` ([which is possible since C#13](https://devblogs.microsoft.com/dotnet/csharp13-calling-methods-is-easier-and-faster/)) to potentially save on array allocation.

If you also introduce a generic extension like this:

```c#
private static void AddTo<T>(this ReadOnlySpan<T> items, ICollection<T> collection)
{
    foreach (var item in items)
        collection.Add(item);
}
```

the adding `arguments` elements to `ArgumentList` of `ProcessStartInfo` would be simple as:

```c#
arguments.AddTo(startInfo.ArgumentList);
```

atifaziz (approved):
Other than a few suggestions, LGTM.

src/CSnakes.Runtime/ProcessUtils.cs
tonybaloney: These commands are run so rarely and take so long I don't think it'll make a noticable difference?

PR #481, opened 2025-06-02
Fix coroutine source generator type order

### LINKED ISSUE #480: `Coroutine[TYield, TSend, TReturn]` is incorrectly reflected as `Task<TYield>` not `Task<TReturn>`
When reflecting the `Coroutine` type, we use the `TYield` type as 

```python
async def example() -> Coroutine[str, int, bytes]:
   x: int = yield "hello"
   return b"sdfsd"
```

Should be reflected as `Task<byte[]>` but it's reflected as `Task<string>`. The tests don't uncover this because we incorrectly put the return type in the `TYield` position.

You can see this when running pyright over the integration tests.

tonybaloney:
Fixes #480 

This is a breaking change


tonybaloney:
It might be better to deprecate this and start another PR to support the simple syntax first

PR #488, opened 2025-06-02
Add missing disposal of buffer exporters in proxies

atifaziz:
This PR adds missing disposal of Python functions returning a _buffer exporter_ in the generated code, which means that they were not being eagerly disposed (they would eventually free up with the next GC cycle).

~~Updating the condition of when the result should be disposed (`resultShouldBeDisposed`) also means that functions returning `None` or `void` no longer dispose the result as `None` is immortal.~~

### DIFF
--- a/src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_basic.approved.txt
+++ b/src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_basic.approved.txt
@@ -274,7 +274,7 @@ public static class TestClassExtensions
             {
                 logger?.LogDebug("Invoking Python function: {FunctionName}", "test_none_result");
                 PyObject __underlyingPythonFunc = this.__func_test_none_result;
-                PyObject __result_pyObject = __underlyingPythonFunc.Call();
+                _ = __underlyingPythonFunc.Call();
                 return;
             }
         }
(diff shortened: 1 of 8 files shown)


src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_basic.approved.txt
```diff
@@ -274,7 +274,7 @@ public static class TestClassExtensions
             {
                 logger?.LogDebug("Invoking Python function: {FunctionName}", "test_none_result");
                 PyObject __underlyingPythonFunc = this.__func_test_none_result;
-                using PyObject __result_pyObject = __underlyingPythonFunc.Call();
+                PyObject __result_pyObject = __underlyingPythonFunc.Call();
```
tonybaloney, line 277: This should be disposed, why is it no longer?

atifaziz: > functions returning `None` or `void` no longer dispose the result as `None` is immortal.

As I pointed out in the PR description, the function signature is known to return `None`, which is immortal, so it's `Dispose` won't do anything anyway.

atifaziz: I'll prepare a separate PR for discarding (see 19b54197c7020f84dea738aa1a40c623accbd0c6). For now, I've scoped this PR to just disposing the _buffer exporter_ on the basis that it's safe to do so because the buffer returned by [`PyObject_GetBuffer`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer) will keep it live:

> On success, fill in _view_, set `view->obj` to a new reference to _exporter_ and return 0.

tonybaloney (approved):

PR #499, opened 2025-06-05
Support PYI type stubs and nested modules

tonybaloney:
The key changes include adding submodule support in CI workflows, implementing a helper class for namespace and class name generation, refining Python static generation logic, and handling edge cases for Python function parsing.

- As a test case the stdlib typeshed submodule was added to `.gitmodules` to support Python type stubs.
- `.pyi` is an acceptable file extension for Python type stubs, and the parser now supports stub functions ending with a colon and ellipsis.
- Users can configure a "root path" which is the root namespace
- Generated modules will use namespaces in both Python and C#

### Namespace and class name generation:
* [`src/CSnakes.SourceGeneration/NamespaceHelper.cs`](diffhunk://#diff-7d98768451292de53bfdb9e7ecbf5b39fe27c96cbb7272cf9a8dd47e281de4a1R1-R75): Introduced a utility class to convert file paths into Python import paths, .NET namespaces, and PascalCase class names, aiding consistent naming conventions and modularity.

### Python static generation improvements:
* [`src/CSnakes.SourceGeneration/PythonStaticGenerator.cs`](diffhunk://#diff-608868082e687f0233b41c8b7391c0d966b5643d37103073d62b6db2adf3b199R11-R47): Enhanced the static generator to support `.pyi` files, added a mechanism to derive namespaces and class names based on configured root directories, and improved logging for module operations. Additionally, implemented error handling for parser crashes and refined diagnostic reporting. 

### Python function parsing:
* [`src/CSnakes.SourceGeneration/Parser/PythonParser.Function.cs`](diffhunk://#diff-f4c4583065ae39abf68ca4c3934c052a1c4890efbbd7a87155b1deed9a008eabR75-R89): Updated the function parser to handle stub definitions ending with a colon and ellipsis, improving compatibility with Python type hinting conventions.

- [x] Add support in parser for stub functions `def foo(): ...`
- [x] Test stdlib stubs
- [x] Add absolute paths to import statements
- [x] Match `__init__.py` as the module root


atifaziz:
As a quick test, I tried changes from PR #524 here and it reduced the number of errors from 565 to 385.

atifaziz (commented):
Sorry, got some minutes to continue the review and didn't get far, so posting some comments meanwhile.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+
+                if (result)
+                {
+                    var methods = ModuleReflection.MethodsFromFunctionDefinitions(functions).ToImmutableArray();
+                    string source = FormatClassFromMethods(@namespace, pascalFileName, methods, moduleAbsoluteName, functions, code, embedSourceSwitch);
+                    sourceContext.AddSource(generatedFileName, source);
+                    sourceContext.ReportDiagnostic(Diagnostic.Create(new DiagnosticDescriptor("PSG002", "PythonStaticGenerator", $"Generated {generatedFileName} from {file.Path}", "PythonStaticGenerator", DiagnosticSeverity.Info, true), Location.None));
+                }
+            } catch (Exception ex) {
+                Location errorLocation = Location.Create(file.Path, TextSpan.FromBounds(0, 1), new LinePositionSpan(new LinePosition(0, 1), new LinePosition(1, 1)));
```
atifaziz, line 88: `errorLocation` is not used anywhere?

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+                sourceContext.ReportDiagnostic(Diagnostic.Create(new DiagnosticDescriptor("PSG999", "PythonStaticGenerator", ex.Message, "PythonStaticGenerator", DiagnosticSeverity.Error, true), null));
+
+#if DEBUG
+                if (!System.Diagnostics.Debugger.IsAttached)
+                {
+                    // This is a good place to debug parser crashes.
+                    // System.Diagnostics.Debugger.Launch();
+                }
+#endif
```
atifaziz, line 97: It this needed? F5 debugging has been available since PR #281.

```suggestion
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
 
-            if (result)
+    /// <summary>
+    /// Get the file path after the relative path.
+    /// .e.g. if the absolute path is "/tmp/foo/bar/baz.py" and the relative path is "foo/bar",
+    /// it will return "baz.py".
+    /// </summary>
+    /// <param name="absolutePath"></param>
+    /// <param name="relativePath"></param>
+    /// <returns></returns>
```
atifaziz, line 109: Remove empty doc elements:

```suggestion
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+    private static string? GetPathAfter(string absolutePath, string relativePath)
+    {
+        // Normalize and split paths
+        var absParts = Path.GetFullPath(absolutePath).Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
+            .Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
+            .ToArray();
```
atifaziz, line 115: Remove extra array conversion (`Split` already returns an array):

```suggestion
            .Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+        var relParts = relativePath.Trim(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
+            .Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
+            .Where(p => !string.IsNullOrEmpty(p))
```
atifaziz, line 118: Why this here and not in the `absPath` case? Is it to normalise adjacent slashes?

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+            .ToArray();
+
+        // Find the index where relParts matches a subsequence in absParts
+        var matchIndex = absParts
+            .Select((_, i) => i)
+            .FirstOrDefault(i => i + relParts.Length <= absParts.Length &&
+                relParts.SequenceEqual(absParts.Skip(i).Take(relParts.Length)));
```
atifaziz, line 125: Comparing spans of the arrays would be quicker:

```suggestion
                relParts.AsSpan().SequenceEqual(absParts.AsSpan(i, relParts.Length)));
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+
+        // If not found, return null
+        if (matchIndex == 0 && (relParts.Length == 0 || !relParts.SequenceEqual(absParts.Take(relParts.Length))))
```
atifaziz, line 128: Comparing spans of the arrays would be quicker:

```suggestion
        if (matchIndex == 0 && (relParts.Length == 0 || !relParts.AsSpan().SequenceEqual(absParts.AsSpan(0, relParts.Length))))
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+                }
+
+                if (result)
+                {
+                    var methods = ModuleReflection.MethodsFromFunctionDefinitions(functions).ToImmutableArray();
+                    string source = FormatClassFromMethods(@namespace, pascalFileName, methods, moduleAbsoluteName, functions, code, embedSourceSwitch);
+                    sourceContext.AddSource(generatedFileName, source);
+                    sourceContext.ReportDiagnostic(Diagnostic.Create(new DiagnosticDescriptor("PSG002", "PythonStaticGenerator", $"Generated {generatedFileName} from {file.Path}", "PythonStaticGenerator", DiagnosticSeverity.Info, true), Location.None));
+                }
+            } catch (Exception ex) {
```
atifaziz, line 87: Generally it is discourages to catch any kind of exception. Can we be more specific? This seems to be a dramatic change of policy and wondering what prompted this.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -58,26 +66,114 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
+        // If not found, return null
+        if (matchIndex == 0 && (relParts.Length == 0 || !relParts.SequenceEqual(absParts.Take(relParts.Length))))
+            return null;
+
+        // Return the parts after the match
+        var afterParts = absParts.Skip(matchIndex + relParts.Length);
+        return string.Join(Path.DirectorySeparatorChar.ToString(), afterParts);
+    }
+
+    public static (string @namespace, string generatedFileName, string pascalFileName, string moduleAbsoluteName) GetNamespaceAndClassName(string path, string configuredRootDir)
```
atifaziz, line 136: A tuple of these many identically typed elements is hard to reason about in terms of order and therefore prone to bugs. I'd like to suggest introducing a record to hold all the names:

```c#
public sealed record DerivedNames
{
    public required string Namespace { get; init; }
    public required string GeneratedFileName { get; init; }
    public required string PascalFileName { get; init; }
    public required string ModuleAbsoluteName { get; init; }
}
```

src/CSnakes.SourceGeneration/NamespaceHelper.cs
```diff
@@ -0,0 +1,75 @@
...
+        string filename = Path.GetFileNameWithoutExtension(path);
+        string directory = Path.GetDirectoryName(path) ?? string.Empty;
+
+        if (string.IsNullOrWhiteSpace(directory))
+        {
+            return filename;
+        }
+
+        var folders = directory.Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
+            .Select(p => p.Replace(".", "_"))
```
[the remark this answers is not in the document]
tonybaloney: If a file is named `foo.bar.py` you can't import it as `foo.bar` (you can't really import it at all actually). I was covering that to ensure it doesn't generate invalid import statements but it'd probably be better to throw an invalid argument exception

```default
> edit foo.bar.py
> python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import foo.bar
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'foo'
```

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
```diff
@@ -8,35 +8,43 @@
...
 
-        context.RegisterSourceOutput(pythonFilesPipeline.Combine(embedPythonSource), static (sourceContext, opts) =>
-        {
-            var file = opts.Left;
-            var embedSourceSwitch = opts.Right;
+        // Get directory traversal root
+        var rootDirectory = context.AnalyzerConfigOptionsProvider.Select(static (options, cancellationToken) =>
+            options.GlobalOptions.TryGetValue("csnakes_root_directory", out var rootDir)
+                ? rootDir
+                : string.Empty); // Default to empty string
```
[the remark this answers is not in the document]
tonybaloney: agreed.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
tonybaloney: Exceptions thrown by a source generator are captured in the build logs but they're hard to find, log, and debug. Also they don't contain any details on which source file caused the exception. You'll see when compiling the stdlib integration tests it still has a few exceptions because there are unsupported syntax features.
 
IF any user gets an error during build because of a bug in the CSnakes source generator we get a better error message as well which they can share in the bug reports.

src/CSnakes.SourceGeneration/PythonStaticGenerator.cs
tonybaloney: That uses some modern C# features we don't have because we're targeting netstandard2.0 but I see your point.

atifaziz: You can light up those modern features via [PolySharp](https://github.com/Sergio0694/PolySharp) (which [we already use](https://github.com/tonybaloney/CSnakes/blob/6496a356dc8b2a720a18808a0ec7f848cad82c88/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj#L19-L22)) by adding the following to the project file:

```diff
diff --git a/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj b/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj
index b7a9bb09..9dd18688 100644
--- a/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj
+++ b/src/CSnakes.SourceGeneration/CSnakes.SourceGeneration.csproj
@@ -30,6 +30,10 @@
       System.Diagnostics.CodeAnalysis.NotNullAttribute;
       System.Diagnostics.CodeAnalysis.NotNullWhenAttribute;
       System.Diagnostics.CodeAnalysis.DoesNotReturnAttribute;
+      System.Diagnostics.CodeAnalysis.SetsRequiredMembersAttribute;
+      System.Runtime.CompilerServices.IsExternalInit;
+      System.Runtime.CompilerServices.RequiredMemberAttribute;
+      System.Runtime.CompilerServices.CompilerFeatureRequiredAttribute;
     </PolySharpIncludeGeneratedTypes>
   </PropertyGroup>
 
```

src/StdLib.Tests/StdLib.Tests.csproj
```diff
@@ -0,0 +1,43 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
+    <PackageReference Include="System.Net.Http" />
+    <PackageReference Include="System.Text.RegularExpressions" />
+    <PackageReference Include="xunit.v3" />
+    <PackageReference Include="xunit.runner.visualstudio" />
+    <PackageReference Include="python" />
```
atifaziz, line 17: Is this needed?

src/StdLib.Tests/StdLib.Tests.csproj
```diff
@@ -0,0 +1,43 @@
...
+
+  <PropertyGroup>
+    <IsPackable>false</IsPackable>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <IsTestProject>true</IsTestProject>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" />
```
atifaziz, line 11: Do we need coverage for this?

src/StdLib.Tests/StdLib.Tests.csproj
```diff
@@ -0,0 +1,43 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" />
+    <PackageReference Include="System.Net.Http" />
+    <PackageReference Include="System.Text.RegularExpressions" />
```
atifaziz, line 14: Are these needed?

src/Integration.Tests/Integration.Tests.csproj
```diff
@@ -29,12 +29,21 @@
   </ItemGroup>
 
   <ItemGroup>
-    <AdditionalFiles Include="python\*.py">
+    <AdditionalFiles Include="python\**\*.py">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </AdditionalFiles>
+    <AdditionalFiles Include="python\**\*.pyi">
```
atifaziz, line 35: I guess this would be something to consider for PR #576 or later?

Issue #454, opened 2025-05-07
Support .pyi type stubs


tonybaloney:
Not every integration needs a Python file. 

We should support generating an interface from a type stub (pyi) file. 

We would want to preserve the path order, eg. 

'urllib/parse.pyi' should be imported as "urllib.parse"

In practice pyi files contain many complex types and Unions that CSnakes does not support. 

In the least we could generate from simple ones. 

Also the overload decorator is common in pyi files and we haven't supported that yet.

PR #502, opened 2025-06-06
Support for `T | None` union as `Optional[T]`

atifaziz:
This PR adds support for modern Python union syntax (`T | None`) as an equivalent to `Optional[T]`, as per [PEP 604](https://peps.python.org/pep-0604/):

> We will also be able to write `t | None` or `None | t` instead of `Optional[t]`

This enhancement modernizes the type parsing capabilities to align with Python 3.10+ union syntax while maintaining backward compatibility with the existing `Optional[T]` notation.

It should partially help with #454 that's being addressed by PR #499.

## Design

The implementation introduces a parsing strategy that recognizes both forms of union syntax:

- `T | None`
- `None | T` (alternative uncommon order)

Both forms are internally converted to the canonical `Optional[T]` representation, ensuring consistent handling throughout the type system.

**The parser deliberately limits union support to the specific case of optional types (`T | None`).** More complex union types (e.g., `int | str | None`) are not yet supported and will result in parse errors, as indicated in the test comments. This focused approach ensures robust handling of the most common union pattern while leaving room for future expansion.

## Test Coverage

- Added unit tests for both `T | None` and `None | T` syntax variants
- Extended signature generation tests to verify C# output correctness
- Added tokenizer tests for pipe character recognition
- Included edge case testing for invalid union combinations
- Updated existing tests to use modern union syntax

## Compatibility

This change is fully backward compatible. Existing codebases using `Optional[T]` syntax will continue to work without modification, while new code can adopt the more modern `T | None` syntax as preferred. The generated C# interfaces remain identical regardless of which Python syntax is used, ensuring seamless interoperability.


tonybaloney:
Python 3.9 goes out of support on October 31. We could drop it shortly after. It'll get increasingly difficult to keep support.

tonybaloney (approved):

PR #505, opened 2025-06-08
Safer & stronger-typed API for buffers

### LINKED ISSUE #504: Buffer API is fundamentally unsafe
Suppose the following Python module with a single test function:

```python
from collections.abc import Buffer

def test(length: int = 10000000, fill: int = 42) -> Buffer:
    return bytearray([42] * length)
```

which is then used from C# as follows:

```csharp
using var test = env.BufferTest();

Span<byte> GetBufferSpan(byte fill)
{
    using var buffer = test.Test(fill: fill);
    return buffer.AsSpan<byte>();
}

const byte fill = 42;
var span = GetBufferSpan(fill);
if (span.IndexOfAnyExcept(fill) is var i and >= 0)
    throw new($"Buffer contains unexpected value at index {i} (expected {fill}, but was {span[i]}).");
```

When run in release configuration, such application code crashes with a `System.AccessViolationException` when attempting to access a Python buffer object's memory:

    Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
    at System.SpanHelpers.NonPackedIndexOfValueType[...]
    at System.SpanHelpers.IndexOfValueType[...]
    at System.SpanHelpers.IndexOfAnyExceptValueType[...]
    at System.MemoryExtensions.IndexOfAnyExcept[...]
    at Program.<Main>$

Environment:

- **OS**: Windows 11
- **Framework**: .NET 9.0
- **Configuration**: Release
- **CSnakes.Runtime**: 1.0.35

The code attempts to access memory from a Python buffer object after the Python object has been disposed. This creates a classic **use-after-free** vulnerability:

```csharp
Span<byte> GetBufferSpan(byte fill)
{
    using var buffer = test.Test(fill: fill);  // ← Python object disposed here
    return buffer.AsSpan<byte>();              // ← Returns span pointing to freed memory
}

const byte fill = 42;
var span = GetBufferSpan(fill);                // ← span now points to invalid memory
if (span.IndexOfAnyExcept(fill) is var i and >= 0)  // ← Access violation occurs here
```

1. Python Object Creation: The Python function creates a `bytearray` with 10 million bytes
3. Buffer Wrapping: CSnakes wraps this in a disposable buffer object
4. Premature Disposal: The `using` statement disposes the buffer at the end of `GetBufferSpan`
5. Invalid Memory Access: The returned `Span<byte>` still references the freed Python memory
6. Crash: When `IndexOfAnyExcept` tries to read the memory, it accesses freed/invalid memory

The fundamental issue here is that CSnakes **allows access to native memory in an unsafe manner without making developers aware of the safety implications**. The current API makes it _too easy_ for unsuspecting developers—particularly those less familiar with buffer and native memory management rules—to make critical mistakes around memory life-time issues.

These mistakes lead to crashes and memory corruption bugs that are:

- Extremely difficult to reproduce (timing-dependent, configuration-dependent)
- Hard to analyze and debug (symptoms may appear far from the root cause)
- Potentially exploitable (memory corruption vulnerabilities)

It may also bypass policies, auditing and linters that would otherwise identify unsafe operations.

While `PyBuffer.AsSpanInternal` is maked `unsafe`:

https://github.com/tonybaloney/CSnakes/blob/058487ad62cad306416039dab059da41970b0418/src/CSnakes.Runtime/Python/PyBuffer.cs#L205-L214

It returns a pointer wrapped in a `Span<T>` that's then returned in the public API that's _not_ marked unsafe:

https://github.com/tonybaloney/CSnakes/blob/058487ad62cad306416039dab059da41970b0418/src/CSnakes.Runtime/Python/PyBuffer.cs#L125

The fix is easy, of course:

```csharp
Span<byte> GetBufferSpan(byte fill)
{
    using var buffer = test.Test(fill: fill);
    var span = buffer.AsSpan<byte>();
    var copy = new byte[span.Length];
    span.CopyTo(copy);
    return copy;
}
```

However, by copying the buffer data into managed memory, it defeats the purpose of returning a buffer in the first place. The whole point of using buffers is to avoid expensive memory allocations and copies, especially for large datasets.

CSnakes should provide clear API boundaries between safe and unsafe operations by (preferably) surfacing the `unsafe` modifier in the public API or reflecting it in the operation name, as in `AsUnsafeSpan`.

## Demonstration Code

- [`demo.zip`](https://github.com/user-attachments/files/20644908/demo.zip) (also [available online as a Gist](https://gist.github.com/atifaziz/823bce2f9197d1873128bdf60b60cfd6))

Run with `dotnet run -c Release` as optimized code execution generally makes the use-after-free more likely to cause an immediate crash.

### [`buffer_test.py`](https://gist.github.com/atifaziz/823bce2f9197d1873128bdf60b60cfd6#file-buffer_test-py)

```python
from collections.abc import Buffer

def test(length: int = 10000000, fill: int = 42) -> Buffer:
    return bytearray([42] * length)
```

### [`Program.cs`](https://gist.github.com/atifaziz/823bce2f9197d1873128bdf60b60cfd6#file-program-cs)

```csharp
using System;
using CSnakes.Runtime;
using CSnakes.Runtime.Locators;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;

var builder = Host.CreateApplicationBuilder();
_ = builder.Services
           .WithPython()
           .FromRedistributable(RedistributablePythonVersion.Python3_12);
builder.Logging.SetMinimumLevel(LogLevel.Debug);
var app = builder.Build();
using var env = app.Services.GetRequiredService<IPythonEnvironment>();
using var test = env.BufferTest();

Span<byte> GetBufferSpan(byte fill)
{
    using var buffer = test.Test(fill: fill);
    return buffer.AsSpan<byte>();
}

const byte fill = 42;
var span = GetBufferSpan(fill);
if (span.IndexOfAnyExcept(fill) is var i and >= 0)
    throw new($"Buffer contains unexpected value at index {i} (expected {fill}, but was {span[i]}).");
```

### [`BufferDemo.csproj`](https://gist.github.com/atifaziz/823bce2f9197d1873128bdf60b60cfd6#file-bufferdemo-csproj)

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net9.0</TargetFramework>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <None Remove="*.py" />
  </ItemGroup>

  <ItemGroup>
    <AdditionalFiles Include="*.py">
    </AdditionalFiles>
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="CSnakes.Runtime" Version="1.0.35" />
  </ItemGroup>

</Project>
```

### Patch

```diff
diff --git a/BufferDemo.csproj b/BufferDemo.csproj
new file mode 100644
index 0000000..d1ec188
--- /dev/null
+++ b/BufferDemo.csproj
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Remove="*.py" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="*.py">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </AdditionalFiles>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CSnakes.Runtime" Version="1.0.35" />
+  </ItemGroup>
+
+</Project>
diff --git a/Program.cs b/Program.cs
new file mode 100644
index 0000000..6fd5385
--- /dev/null
+++ b/Program.cs
@@ -0,0 +1,26 @@
+using System;
+using CSnakes.Runtime;
+using CSnakes.Runtime.Locators;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+var builder = Host.CreateApplicationBuilder();
+_ = builder.Services
+           .WithPython()
+           .FromRedistributable(RedistributablePythonVersion.Python3_12);
+builder.Logging.SetMinimumLevel(LogLevel.Debug);
+var app = builder.Build();
+using var env = app.Services.GetRequiredService<IPythonEnvironment>();
+using var test = env.BufferTest();
+
+Span<byte> GetBufferSpan(byte fill)
+{
+    using var buffer = test.Test(fill: fill);
+    return buffer.AsSpan<byte>();
+}
+
+const byte fill = 42;
+var span = GetBufferSpan(fill);
+if (span.IndexOfAnyExcept(fill) is var i and >= 0)
+    throw new($"Buffer contains unexpected value at index {i} (expected {fill}, but was {span[i]}).");
diff --git a/buffer_test.py b/buffer_test.py
new file mode 100644
index 0000000..0826d74
--- /dev/null
+++ b/buffer_test.py
@@ -0,0 +1,4 @@
+from collections.abc import Buffer
+
+def test(length: int = 10000000, fill: int = 42) -> Buffer:
+    return bytearray([42] * length)
```

atifaziz:
This PR is a proposal to address and close #504. It refactors the buffer API to avoid critical memory safety vulnerabilities that can easily lead to `System.AccessViolationException` crashes when juggling the lifetime management rules of native memory backing Python buffer objects, and spans that may be pointing into them. This is done by making most of the API use safe patterns, where spans are provided as inputs to functions as opposed to being handed out, e.g.:

```csharp
public interface PyArrayBuffer<T> : IPyBuffer<T>, IMemoryOwner<T> where T : unmanaged
{
    public T this[int index]
    {
        get => UnsafeAsReadOnlySpan()[index];
        set => UnsafeAsSpan()[index] = value;
    }

    public TResult Map<TResult>(ReadOnlySpanFunc<T, TResult> function) =>
        function(UnsafeAsReadOnlySpan());

    public TResult Map<TArg, TResult>(in TArg arg, ReadOnlySpanFunc<T, TArg, TResult> function)
        where TArg : allows ref struct =>
        function(UnsafeAsReadOnlySpan(), arg);

    public TResult Map<TArg1, TArg2, TResult>(in TArg1 arg1, in TArg2 arg2, ReadOnlySpanFunc<T, TArg1, TArg2, TResult> function)
        where TArg1 : allows ref struct
        where TArg2 : allows ref struct =>
        function(UnsafeAsReadOnlySpan(), arg1, arg2);

    public TResult Map<TArg1, TArg2, TArg3, TResult>(in TArg1 arg1, in TArg2 arg2, in TArg3 arg3, ReadOnlySpanFunc<T, TArg1, TArg2, TArg3, TResult> function)
        where TArg1 : allows ref struct
        where TArg2 : allows ref struct
        where TArg3 : allows ref struct =>
        function(UnsafeAsReadOnlySpan(), arg1, arg2, arg3);

    public void Do(SpanAction<T> action) =>
        action(UnsafeAsSpan());

    public void Do<TArg>(in TArg arg, SpanAction<T, TArg> action)
        where TArg : allows ref struct =>
        action(UnsafeAsSpan(), arg);

    public void Do<TArg1, TArg2>(in TArg1 arg1, in TArg2 arg2, SpanAction<T, TArg1, TArg2> action)
        where TArg1 : allows ref struct
        where TArg2 : allows ref struct =>
        action(UnsafeAsSpan(), arg1, arg2);

    public void Do<TArg1, TArg2, TArg3>(in TArg1 arg1, in TArg2 arg2, in TArg3 arg3, SpanAction<T, TArg1, TArg2, TArg3> 
        where TArg1 : allows ref struct
        where TArg2 : allows ref struct
        where TArg3 : allows ref struct =>
        action(UnsafeAsSpan(), arg1, arg2, arg3);

    public void CopyFrom(scoped in ReadOnlySpan<T> source) => source.CopyTo(UnsafeAsSpan());
    public void CopyTo(scoped in Span<T> destination) => UnsafeAsReadOnlySpan().CopyTo(destination);
    public Span<T> UnsafeAsSpan() => UnsafeAsSpan(writeable: true);
    public ReadOnlySpan<T> UnsafeAsReadOnlySpan() => UnsafeAsSpan(writeable: false);

    /// <summary>
    /// Gets the memory directly underlying the buffer, which is tied to the lifetime of the buffer.
    /// <em>Usage after disposing the buffer will lead to corruption and crashes</em>.
    /// </summary>
    /// <remarks>
    /// See <see
    /// href="https://learn.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines"><see
    /// cref="Memory{T}"/> and <see cref="Span{T}"/> usage guidelines</see> for more information.
    /// </remarks>
    public Memory<T> UnsafeMemory => (_memoryManager ??= new UnmanagedMemoryManager(this)).Memory;
}
```

The only unsafe access is provided by `UnsafeMemory`, which returns a `Memory<T>` (as being discussed in #471), and `UnsafeAsSpan` and `UnsafeAsReadOnlySpan`. Ideally, these [would be annotated as `unsafe`](https://github.com/dotnet/designs/blob/566ad4cafcc578d6389c215c61924ee9e07dcb29/accepted/2025/memory-safety/caller-unsafe.md), but the modifier won't have the desired effect so it's spelled out in the name of the members considered dangerous.

```csharp
// Before: Dangerous pattern that caused crashes
Span<byte> GetBufferSpan(byte fill)
{
    using var buffer = test.Test(fill: fill);  // ⬅️ Disposed here
    return buffer.AsSpan<byte>();              // ❌ Dangling pointer
}

// After: Type-safe access with explicit lifetime management
if (buffer is IPyArrayBuffer<byte> typedBuffer)
{
    // Safe: Access within buffer lifetime
    var result = typedBuffer.Map(span => span.IndexOfAnyExcept(fill));
    // or use IMemoryOwner<T> pattern for explicit lifetime control
}
```

I've made the API more type-safe:

- Before, there was a single `PyBuffer` class with run-time type checking and casting
- Now there's a generic type-safe hierarchy with:
  - `PyBuffer<T>` representing a shape-less base for some buffer of `T` items
  - `PyArrayBuffer<T>` for scalars and 1D arrays
  - `PyArray2DBuffer<T>` for 2D arrays
  - `PyTensorBuffer<T>` for tensors

This eliminates run-time type mismatches and provides compile-time safety. Developers can now rely on standard C# patterns like:

```csharp
// Downcasting
var byteBuffer = (PyArrayBuffer<byte>)buffer;

// Safe casting (with simple pattern matching) like with 'as'
if (buffer is PyArray2DBuffer<float> floatBuffer2D)
{
    // Work with 2D float buffer
}

// Pattern matching
switch (buffer)
{
    case PyArrayBuffer<int> intBuffer: /* handle 1D int array */
    case PyArray2DBuffer<double> doubleBuffer2D: /* handle 2D double array */
}
```

It also improves performance as all checks and creation of the right buffer sub-type is done once. The validation also lies with each sub-type where it makes sense.

## Other Notes

- Buffers now properly handle disposal in both managed and finalizer contexts. Like in the case of `PyObject`, added buffer-specific disposal queue that safely handles cleanup when the GIL isn't available (see also #544).
- Deprecated all existing `AsSpan<T>()` family of methods to discourage unsafe usage patterns.
- A host of new delegates (see `SpanDelegates.cs`) had to be introduced for `Do` and `Map` family of methods of each buffer type.
- This proposal introduces breaking changes.

## Pending Decisions & Work

This is a draft PR to summon early feedback.

- [x] Decide on `IPyBuffer` sub-type/hierarchical API
- [x] Decide on interfaces with default methods vs classes for buffer types
- [x] Decide on safe API in addition to `Map`, `Do`, `CopyTo`, `CopyFrom`, etc. (including overloads)
- [x] Decide on unsafe API
- [ ] Add benchmarks for performance impact of using a safe indexer?
- [x] Update documentation (`buffers.md`)
- [ ] Ensure thread-safe disposal & check


src/CSnakes.Runtime/CPython/Buffer.cs
```diff
@@ -17,20 +17,27 @@ private enum PyBUF : int
...
+        public nint  /* PyObject*   */ obj;
+        public nint  /* Py_ssize_t  */ len;
+        public nint  /* Py_ssize_t  */ itemsize;
+        public int   /* int         */ @readonly;
+        public int   /* int         */ ndim;
+        public nint  /* char*       */ format;
+        public nint* /* Py_ssize_t* */ shape;
+        public nint* /* Py_ssize_t* */ strides;
+        public nint* /* Py_ssize_t* */ suboffsets;
+        public nint* /* void*       */ @internal;
```
AaronRobinsonMSFT:
```suggestion
        public nint  /* void*       */ @internal;
```

Or ideally `void*`. The `nint*` is clearly wrong though.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -1,62 +1,30 @@
...
-internal sealed class PyBuffer : IPyBuffer
+
+public interface IPyBuffer<T> : IPyBuffer where T : unmanaged;
+
+public class PyBuffer<T> : IPyBuffer<T> where T : unmanaged
 {
-    private CPythonAPI.Py_buffer _buffer;
-    private readonly string _format;
-    private readonly ByteOrder _byteOrder;
+    private protected static readonly int ItemSize = Unsafe.SizeOf<T>();
```
AaronRobinsonMSFT:
```suggestion
    private protected static readonly int ItemSize = sizeof(T);
```

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -109,231 +77,225 @@ private ref readonly CPythonAPI.Py_buffer Buffer
         }
     }
 
-    public long Length => Buffer.len;
+    private protected unsafe ref T TypedRef => ref Unsafe.AsRef<T>(Pointer);
```
AaronRobinsonMSFT: This API is a bit suspicious with the `Pointer` below. It muddles how people should be thinking about this fields ownership. Either it is a raw pointer or it is a fully managed reference. Exposing both is confusing to me as a consumer. I would defer to providing a pointer and letting people know they need to think about the pointer value properly rather than exposing a `ref T` version that could be abused.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -109,231 +77,225 @@ private ref readonly CPythonAPI.Py_buffer Buffer
...
-        if (_format.Length == 0)
+    public Type ItemType => typeof(T);
+
+    private protected unsafe ReadOnlySpan<nint> Shape =>
+        Buffer switch
         {
-            return ByteOrder.Native;
-        }
+            { shape: not null and var shape, strides: not null, ndim: var ndim } => new ReadOnlySpan<nint>(shape, ndim),
+            _ => throw new InvalidOperationException("Buffer does not have shape and strides."),
```
AaronRobinsonMSFT: This is a bit odd to me. You're not using or expressing stride here, but you require it. The returned type `ReadOnlySpan<nint>` doesn't capture stride or express it, so why is that a requirement? Seems more appropriate to have that check as the use site.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -109,231 +77,225 @@ private ref readonly CPythonAPI.Py_buffer Buffer
...
-    private unsafe void Validate2DBufferCommon<T>() where T : unmanaged
-    {
-        EnsureDimensions(2);
-        EnsureShapeAndStrides();
-        if (Shape[0] * Shape[1] * sizeof(T) != Length)
-        {
-            throw new InvalidOperationException("Buffer length is not equal to shape");
-        }
-    }
+        var format = Marshal.PtrToStringUTF8(buffer.format) ?? "B";
```
AaronRobinsonMSFT: I would defer to the [`Utf8StringMarshaller`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshalling.utf8stringmarshaller) if possible.

src/CSnakes.Runtime/Python/PyArrayBuffer.cs
```diff
@@ -0,0 +1,145 @@
+using CSnakes.Runtime.CPython;
+using System.Buffers;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace CSnakes.Runtime.Python;
+
+public sealed class PyArrayBuffer<T> : PyBuffer<T>, IMemoryOwner<T> where T : unmanaged
```
AaronRobinsonMSFT, line 7: An alternative approach would be to use a `SafeHandle` abstraction and avoid the racey and ownership confusion that can come with implementing the `IDisposable` pattern.

src/CSnakes.Runtime/Python/PyBuffer.cs
tannergooding: This should notably become safe in C# 15 as well, as part of the `unsafe evolution` feature work.

src/CSnakes.Runtime/CPython/Buffer.cs
tannergooding: Similar comment for other cases like `buf` (`void*`), `obj` (`PyObject*`), and `format` (`char*`).

For `format` in particular, the `char*` should be `byte*` or `sbyte*`. -- In C, `char` is signed by default on most ABIs but .NET tends to use `byte` for UTF-8 strings, so depending on which is more appropriate. It doesn't map to C# `char` since that's 2-bytes, not 1.

atifaziz: I actually did not make any changes to the definition of this struct in this PR except adding the [corresponding C field types](https://github.com/python/cpython/blob/6280bb547840b609feedb78887c6491af75548e8/Include/pybuffer.h#L20-L33) as comments so that I don't have to keep switching back and forth when re-modelling. I was going to bring this in as a separate PR, but I think I might as well do that now to address some of the suggestions.

> Or ideally `void*`. The `nint*` is clearly wrong though.

@AaronRobinsonMSFT That's definitely wrong! I'll fix that separately as its unrelated to this PR.

atifaziz: Addressed in PR #830.

Thanks, @AaronRobinsonMSFT & @tannergooding.

src/CSnakes.Runtime/Python/PyBuffer.cs
atifaziz: @AaronRobinsonMSFT, I think as per @tannergooding's comment, this is not possible today. It gives error [CS0233](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/unsafe-code-errors#pointer-types-and-managed-types):

> 'T' does not have a predefined size, therefore sizeof can only be used in an unsafe context 

For now, I've added a to-do comment with be48e7170ba39412e7f801e5e248d8a6de04defe to revisit this when C# 15 lands.

src/CSnakes.Runtime/Python/PyBuffer.cs
atifaziz: Addressed via PR #830 and merged here with 7484f6a7968932fcc3d76c7b61df0dc65cbfdb38 ([although it seems that `Utf8StringMarshaller.ConvertToManaged` defers to `Marshal.PtrToStringUTF8`](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/Utf8StringMarshaller.cs#L42) anyway).

src/CSnakes.Runtime/Python/PyBuffer.cs
atifaziz: This is a really good point and removed with a8b0e5619bf8a9be2eb1e5fac7ad95c0d2c1d26c.

src/CSnakes.Runtime/Python/PyBuffer.cs
atifaziz: This PR was about refactoring the public API surface to encourage memory-safe usage, via scoped usage of spans on unmanaged memory behind the buffers. I took special care not to revisit internals in order to avoid a sea of red & green changes that would make things hard to review. Most internal changes were therefore mechanical movements. In this specific case, the code was like this already. The shape and strides requirement, including the exception, was coming from the validation method `EnsureShapeAndStrides`:

https://github.com/tonybaloney/CSnakes/blob/ff41f263bc95560504fc029dabae7adbc505f720/src/CSnakes.Runtime/Python/PyBuffer.cs#L207-L213

I in-lined this since there was only one call site remaining (instead of 4 previously) after the refactoring:

https://github.com/tonybaloney/CSnakes/blob/ff41f263bc95560504fc029dabae7adbc505f720/src/CSnakes.Runtime/Python/PyBuffer.cs#L121-L128

I agree that `strides` being required but not used _for shape_ is odd, but as I said, I was trying to refactor internals very mechanically.

Technically, this check is very defensive. It would mean a violation of [Python's buffer protocol](https://docs.python.org/3/c-api/buffer.html) by the _exporter_. CSnakes always [requests the buffer](https://docs.python.org/3/c-api/buffer.html#contiguity-requests) with the [`PyBUF_C_CONTIGUOUS`](https://docs.python.org/3/c-api/buffer.html#c.PyBUF_C_CONTIGUOUS) flag and the exporter _must_ fill `shape` and `strides` fields of `Py_buffer`. In fact, [`PyBUF_C_CONTIGUOUS` includes `PyBUF_STRIDES`](https://github.com/python/cpython/blob/776573c9f08f70ae9cb2327a905a42f265b331a6/Include/pybuffer.h#L119) so it's impossible to bypass per the protocol:

```c
#define PyBUF_C_CONTIGUOUS (0x0020 | PyBUF_STRIDES)
```

The doc for [`PyObject_GetBuffer`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer) also states :

> Send a request to _exporter_ to fill in _view_ as specified by _flags_. If the exporter cannot provide a buffer of the exact type, it MUST raise [`BufferError`](https://docs.python.org/3/library/exceptions.html#BufferError), set `view->obj` to `NULL` and return `-1`.

So all in all, this check is very defensive against an untrusted exporter. The check could be done much earlier, but I'll defer this to @tonybaloney and another PR if it needs revisiting.

src/CSnakes.Runtime/Python/PyBuffer.cs
AaronRobinsonMSFT: That doesn't follow for me. This class is already "unsafe" because it has a pointer property below. Remove the "unsafe" from the property move it to the type and avoid the `Unsafe` API call. You're already constrained the types to to be `unmanaged` so `sizeof` will just work. At this point we're just fiddling with obnoxious and unhelpful compiler warnings. I don't see CS0233 as a meaningful blocker for this, `sizeof(T)` is precisely what you want to have here.

src/CSnakes.Runtime/Python/PyBuffer.cs
AaronRobinsonMSFT: Sounds good. Let's add a comment here as to why this may or may not need to be enforced. It sounds like it is an invariant that we could permit being an assert, but have as a hard check for sanity. Commenting on that would be helpful.

src/CSnakes.Runtime/Python/PyBuffer.cs
tannergooding: Right, I was rather noting that this will require:
```csharp
private protected static unsafe readonly int ItemSize = sizeof(T);
```

And in C# 15 that should be relaxed so it can become just:
```csharp
private protected static readonly int ItemSize = sizeof(T);
```

atifaziz: Ah, sorry, I misunderstood. Actually, it never even occurred to me to try and put `unsafe` as a static field modifier when I got the error!

Done with afee312c989366d047adfe390571783b48f687eb.

src/CSnakes.Runtime/PublicAPI/net8.0/PublicAPI.Unshipped.txt
```diff
@@ -6,7 +6,65 @@ CSnakes.Runtime.Python.IAwaitable.WaitAsync(System.Threading.CancellationToken c
...
+CSnakes.Runtime.Python.PyArray2DBuffer<T>.CopyFrom(scoped in CommunityToolkit.HighPerformance.ReadOnlySpan2D<T> source) -> void
+CSnakes.Runtime.Python.PyArray2DBuffer<T>.CopyTo(scoped in CommunityToolkit.HighPerformance.Span2D<T> destination) -> void
+CSnakes.Runtime.Python.PyArray2DBuffer<T>.CopyTo(scoped in System.Span<T> destination) -> void
+CSnakes.Runtime.Python.PyArray2DBuffer<T>.this[int row, int column].get -> T
+CSnakes.Runtime.Python.PyArray2DBuffer<T>.this[int row, int column].set -> void
+CSnakes.Runtime.Python.PyArrayBuffer<T>
+CSnakes.Runtime.Python.PyArrayBuffer<T>.CopyFrom(scoped in System.ReadOnlySpan<T> source) -> void
+CSnakes.Runtime.Python.PyArrayBuffer<T>.CopyTo(scoped in System.Span<T> destination) -> void
+CSnakes.Runtime.Python.PyArrayBuffer<T>.Map<TArg, TResult>(in TArg arg, CSnakes.Runtime.ReadOnlySpanFunc<T, TArg, TResult>! function) -> TResult
+CSnakes.Runtime.Python.PyArrayBuffer<T>.Map<TResult>(CSnakes.Runtime.ReadOnlySpanFunc<T, TResult>! function) -> TResult
```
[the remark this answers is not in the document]
atifaziz: This is a false negative. The other overloads are present on 23 & 24 and the analyzer would have failed the build if they were missing.

src/CSnakes.Runtime/Python/PyBufferExtensions.cs
```diff
@@ -4,114 +4,197 @@
...
     public static TensorSpan<float> AsFloatTensorSpan(this IPyBuffer buffer) => buffer.AsTensorSpan<float>();
+    [Obsolete($"Cast to '{nameof(PyArrayBuffer<>)}<double>' and use its '{nameof(PyArrayBuffer<>.Map)}' method instead.")]
     public static TensorSpan<double> AsDoubleTensorSpan(this IPyBuffer buffer) => buffer.AsTensorSpan<double>();
+    [Obsolete($"Cast to '{nameof(PyArrayBuffer<>)}<nint>' and use its '{nameof(PyArrayBuffer<>.Map)}' method instead.")]
     public static TensorSpan<nint> AsIntPtrTensorSpan(this IPyBuffer buffer) => buffer.AsTensorSpan<nint>();
+    [Obsolete($"Cast to '{nameof(PyArrayBuffer<>)}<nuint>' and use its '{nameof(PyArrayBuffer<>.Map)}' method instead.")]
     public static TensorSpan<nuint> AsUIntPtrTensorSpan(this IPyBuffer buffer) => buffer.AsTensorSpan<nuint>();
+    #endregion
 
+    #region AsTensorSpan
```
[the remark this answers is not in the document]
atifaziz: Fixed with 367c5927c65bcff094969f9c3f4b034da29617d5.

src/CSnakes.Runtime/Python/PyArrayBuffer.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// Returns a span <em>directly</em> over the buffer.
+    /// <em>Usage after disposing the buffer will lead to corruption and crashes</em>.
+    /// </summary>
+    public Span<T> UnsafeAsSpan() => UnsafeAsSpan(writeable: true);
+
+    /// <summary>
+    /// Returns a read-only span <em>directly</em> over the buffer.
+    /// <em>Usage after disposing the buffer will lead to corruption and crashes</em>.
+    /// </summary>
+    /// <exception cref="InvalidOperationException">Underlying buffer is read-only.</exception>
```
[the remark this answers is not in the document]
atifaziz: Fixed with a69503e1fd72c7bd1563fdab241652536bb22523.

src/CSnakes.Runtime/Python/PyArray2DBuffer.cs
```diff
@@ -0,0 +1,125 @@
...
+#endif
+        => action(UnsafeAsSpan2D(), arg1, arg2);
+
+    public void Do<TArg1, TArg2, TArg3>(TArg1 arg1, TArg2 arg2, TArg3 arg3, Span2DAction<T, TArg1, TArg2, TArg3> action)
+#if NET9_0_OR_GREATER
+        where TArg1 : allows ref struct
+        where TArg2 : allows ref struct
+        where TArg3 : allows ref struct
+#endif
+        => action(UnsafeAsSpan2D(), arg1, arg2, arg3);
```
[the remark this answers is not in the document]
atifaziz: Fixed with 05a4100588b9835d7e50d81f4f3649b227034f42.

src/CSnakes.Runtime/Python/PyArrayBuffer.cs
```diff
@@ -0,0 +1,146 @@
+using CSnakes.Runtime.CPython;
+using System.Buffers;
+using System.Reflection;
```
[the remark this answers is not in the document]
atifaziz: Was already removed with a406c482baffca33124ba7ad439750d02d6a4640.

src/CSnakes.Runtime/Python/IPyBuffer.cs
```diff
@@ -24,19 +24,36 @@ public interface IPyBuffer : IDisposable
...
+    [Obsolete($"Cast to '{nameof(PyArray2DBuffer<>)}' and use its '{nameof(PyArrayBuffer<>.Map)}' method instead.")]
+    ReadOnlySpan2D<T> AsReadOnlySpan2D<T>() where T : unmanaged;
 
 #if NET9_0_OR_GREATER
+    [Obsolete($"Cast to '{nameof(PyTensorBuffer<>)}' and use its '{nameof(PyArrayBuffer<>.Do)}' method instead.")]
     TensorSpan<T> AsTensorSpan<T>() where T : unmanaged;
-    ReadOnlyTensorSpan<T> AsReadOnlyTensorSpan<T>() where T : unmanaged;
-#endif
 
+    [Obsolete($"Cast to '{nameof(PyTensorBuffer<>)}' and use its '{nameof(PyArrayBuffer<>.Map)}' method instead.")]
```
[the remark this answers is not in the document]
atifaziz: Fixed with 19a568f9ceef42648cc0015cd9a0c9f55f5ff7fb.

src/CSnakes.Runtime/Python/PyBuffer.cs
```diff
@@ -110,231 +78,224 @@ private ref readonly CPythonAPI.Py_buffer Buffer
...
-    {
-        EnsureDimensions(2);
-        EnsureShapeAndStrides();
-        if (Shape[0] * Shape[1] * sizeof(T) != Length)
-        {
-            throw new InvalidOperationException("Buffer length is not equal to shape");
-        }
-    }
+        string format;
+        unsafe { format = Utf8StringMarshaller.ConvertToManaged(buffer.format) ?? "B"; }
```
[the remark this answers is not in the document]
atifaziz: It's the right call per [`format` documentation](https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.format) in the [Buffer protocol](https://docs.python.org/3/c-api/buffer.html):

> If this is `NULL`, `"B"` (unsigned bytes) is assumed.

src/CSnakes.Runtime/Python/IPyBuffer.cs
```diff
@@ -24,19 +24,36 @@ public interface IPyBuffer : IDisposable
...
+    ReadOnlyTensorSpan<T> AsReadOnlyTensorSpan<T>() where T : unmanaged;
```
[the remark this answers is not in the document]
atifaziz: Fixed with 47f63e7a4011dcc1ff106a811d18c5aa589e1018).

src/CSnakes.Runtime/Python/PyBuffer.cs
atifaziz: With 48dac354348befc1e5b1eb40fefa562f5cf2e801, I have:

- moved the check earlier and in one place since it applies to all types of buffers
- added debug assertions
- added comments explaining why shape and strides are required for all buffers

PR #523, opened 2025-06-13
Fix multi line trailing comma not tokenizing as a closing parenthesis

### LINKED ISSUE #522: Trailing comma in multiline function definition fails parser
The following code fails to parse because of the trailing comma: 

```python
def a(
    opener: str = '',
) -> Any:
    pass
```

tonybaloney:
Fixes #522 

This PR changes how the function definitions are tokenized. Previously each line was tokenized and added to a buffer until a "end of function definition" token was spotted, then the buffered tokens were parsed. This causes issues for tokens which are spread across multiple lines, such as `, )` and `, **`.

Code formatted using black, ruff and other formatters will force code into multiple lines so we have seen some valid code fail to parse because of the way it was being tokenized.

To get around this, the function definition parser is changed to re-tokenize the buffer before parsing. This will alleviate this issue and probably others.


tonybaloney:
This bug probably applies to other tokens like `CommaSlash`, `CommaStar` when separated by a line break since we tokenize them separately then merge the token buffers together to parse them.

We could either:
 - Change the function parser to re-tokenize the merged lines instead of just concatenating them and strip out the line breaks
 - Add variations to the parser, e.g. `Comma` , `Star`

tonybaloney:
Decided on option 1.

atifaziz (approved):
This is fine as a hack to make it work for, but I'm afraid more work will be needed still for comments. For example, this doesn't parse:

```python
def a(
    opener: str = 'foo', # comment
) -> Any:
    pass
```

There are instances of those in the stubs, e.g. see [`stubs/toml/toml/decoder.pyi`](https://github.com/python/typeshed/blob/ecd5141cc036366cc9e3ca371096d6a14b0ccd13/stubs/toml/toml/decoder.pyi#L30-L41).

It's also not uncommon to have [`# type: ignore`](https://typing.python.org/en/latest/spec/directives.html#type-ignore-comments) class of comments next to arguments.

PR #526, opened 2025-06-14
Upgrade to Aspire 9.3 and dependency cleanup

buvinghausen:
Removed all references to System.Text.Json as it is provided by the .net9.0 runtime
Removed unused references from weather.py
Removed all references that are inherited from project dependencies
Changed all dependences to floating at the minor level but Aspire dependencies float at the patch level

The reason for the latter is the AppHost project must have the Aspire SDK listed in full ex: 9.3.0 and so we only want patches within that release cycle.


tonybaloney:
thanks for the PR!

Issue #537, opened 2025-06-16
Parse elipsis as undefined default constant


tonybaloney:
Child of #527 

```
def foo(x: str = ...) 
```

Should mean "x is a string, but the default value is undefined". This is common in type stubs.

We would potentially reflect it as a `string? x = null`, meaning the default would be later specified by the Python function.

PR #540, opened 2025-06-16
Add `...` as a possible default value for use in type stubs

tonybaloney:
Implements #537

### DIFF
--- a/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
+++ b/src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
@@ -22,7 +22,7 @@ public static partial class PythonParser
             from rest in UnderScoreOrDigit.Or(Character.In('.', 'e', 'E', '+', '-')).IgnoreMany()
             select Unit.Value;
 
-    public static TextParser<Unit> HexidecimalConstantToken { get; } =
+    public static TextParser<Unit> HexadecimalConstantToken { get; } =
         from prefix in Span.EqualTo("0x")
         from digits in Character.EqualTo('_').Or(Character.HexDigit).AtLeastOnce()
         select Unit.Value;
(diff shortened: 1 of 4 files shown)


src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
```diff
@@ -74,7 +74,7 @@ from close in Character.EqualTo('\'')
     public static TokenListParser<PythonToken, PythonConstant.HexidecimalInteger> HexidecimalIntegerConstantTokenizer { get; } =
         Token.EqualTo(PythonToken.HexidecimalInteger)
         .Select(d => new PythonConstant.HexidecimalInteger(long.Parse(d.ToStringValue().Substring(2).Replace("_", ""), NumberStyles.HexNumber)))
```
tonybaloney, line 76: This misspelling is all over the place, will fix in another PR

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -50,6 +50,7 @@ public class GeneratedSignatureTests
     [InlineData("def hello(a: Union[int, str] = 5) -> Any:\n ...\n", "PyObject Hello(PyObject? a = null)")]
     [InlineData("def hello(data: Literal[1, 'two', 3.0]) -> None:\n ...\n", "void Hello(PyObject data)")]
     [InlineData("def hello(n: None = None) -> None:\n ...\n", "void Hello(PyObject? n = null)")]
+    [InlineData("def hello(n: Foo = ...) -> None:\n ...\n", "void Hello(PyObject? n = null)")]
```
atifaziz, line 53: `n: Foo = ...` should translate to `PyObject n`. I can understand if `n` was type-annotated as `Foo | None`, but `Foo` says that `None` isn't expected.

src/CSnakes.SourceGeneration/Parser/PythonParser.Constants.cs
atifaziz: So revert it in this PR?

PR #552, opened 2025-06-21
Fix bug in TensorSpan<T> and ReadOnlyTensorSpan<T> Length correctly as item length not byte length

### LINKED ISSUE #550: Bug: TensorSpan initialized with byte length instead of element count in PyBuffer
In the `PyBuffer` class, there's a bug in both `AsTensorSpanInternal<T>` and `AsReadOnlyTensorSpanInternal<T>` methods where the `TensorSpan`/`ReadOnlyTensorSpan` is being initialized with the raw byte length instead of the element count.

### Current Implementation
```csharp
return new TensorSpan<T>(
    (T*)buffer.buf,
    buffer.len,  // This is passing the byte length as dataLength
    Shape,
    strides
);
```

### Problem
The `buffer.len` value represents the total number of bytes in the buffer, not the number of elements. This is incorrect because `TensorSpan<T>`'s constructor [expects the `dataLength` parameter to be the number of elements](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.tensors.tensorspan-1.-ctor?view=net-9.0-pp#system-numerics-tensors-tensorspan-1-ctor(-0*-system-intptr-system-readonlyspan((system-intptr))-system-readonlyspan((system-intptr)))), not bytes.

### Impact
This causes the TensorSpan to be initialized with an incorrect length, which could lead to:
- Incorrect memory access
- Potential buffer overflows
- Wrong calculations when using the tensor

### Fix
The `dataLength` should be calculated by dividing the byte length by the size of type T:

```csharp
return new TensorSpan<T>(
    (T*)buffer.buf,
    buffer.len / sizeof(T),  // Convert byte length to element count for dataLength
    Shape,
    strides
);
```

The same fix needs to be applied to both `AsTensorSpanInternal<T>` and `AsReadOnlyTensorSpanInternal<T>` methods.

tonybaloney:
Fixes #550

### DIFF
--- a/src/CSnakes.Runtime/Python/PyBuffer.cs
+++ b/src/CSnakes.Runtime/Python/PyBuffer.cs
@@ -2,6 +2,8 @@ using CommunityToolkit.HighPerformance;
 using CSnakes.Runtime.CPython;
 using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.Marshalling;
+using System.Diagnostics;
+
 #if NET9_0_OR_GREATER
 using System.Numerics.Tensors;
 #endif
@@ -265,7 +267,7 @@ internal sealed class PyBuffer : IPyBuffer
         }
         return new TensorSpan<T>(
             (T*)_buffer.buf,
-            _buffer.len,
+            _buffer.len / sizeof(T),
             Shape,
             strides
         );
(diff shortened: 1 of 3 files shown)


tonybaloney:
CC @phizch

PR #569, opened 2025-06-25
Implement Unions as overloads using the product of the combinations

tonybaloney:


src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs
```diff
@@ -13,62 +13,76 @@ public enum ConversionDirection
...
         {
             ({ Name: "list" or "typing.List" or "List" or "typing.Sequence" or "Sequence", Arguments: [var t] }, _) => CreateListType(t, direction),
             ({ Name: "tuple" or "typing.Tuple" or "Tuple", Arguments: var ts }, _) => CreateTupleType(ts, direction),
             ({ Name: "dict" or "typing.Dict" or "Dict" or "typing.Mapping" or "Mapping", Arguments: [var kt, var vt] }, _) => CreateDictionaryType(kt, vt, direction),
-            ({ Name: "typing.Optional" or "Optional", Arguments: [var t] }, _) => SyntaxFactory.NullableType(AsPredefinedType(t, direction)),
+            ({ Name: "typing.Optional" or "Optional", Arguments: [var t] }, _) => AsPredefinedType(t, direction).Select(SyntaxFactory.NullableType),
             ({ Name: "typing.Generator" or "Generator", Arguments: [var yt, var st, var rt] }, _) => CreateGeneratorType(yt, st, rt, direction),
             ({ Name: "typing.Coroutine" or "Coroutine", Arguments: [var yt, var st, var rt] }, _) => CreateCoroutineType(yt, st, rt, direction),
+            ({ Name: "typing.Union" or "Union", Arguments: var ts }, _) => [.. ts.SelectMany(t => AsPredefinedType(t, direction))],
```
tonybaloney, line 25: This should not happen for both directions, only to Python

src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameter.cs
```diff
@@ -97,45 +97,145 @@ [.. Regular.Select(regularProjector)],
...
+                pos--;
+            }
+            if (pos < 0)
+                break;
         }
+    }
 
-        StringBuilder Delimit() => sb.Length > 1 ? sb.Append(", ") : sb;
+    public IEnumerable<PythonFunctionParameterList<TResult>>
+        MapMany<TResult>(Func<T, IEnumerable<TResult>> positionalProjector,
```
atifaziz, line 133: I feel this implementation could be simplified a lot by taking advantage of existing methods of the `PythonFunctionParameterList<T>` that were designed to avoid having to deal with the various parameter groups:

```c#
    enum ParameterGroup
    {
        Positional,
        Regular,
        VariadicPositional,
        Keyword,
        VariadicKeyword
    }

    public IEnumerable<PythonFunctionParameterList<TResult>>
        MapMany<TResult>(Func<T, IEnumerable<TResult>> positionalProjector,
                         Func<T, IEnumerable<TResult>> regularProjector,
                         Func<T, IEnumerable<TResult>> variadicPositionalProjector,
                         Func<T, IEnumerable<TResult>> keywordProjector,
                         Func<T, IEnumerable<TResult>> variadicKeywordProjector)
        where TResult : class
    {
        /* Each parameter group will be a list of parameters with a list of potential types.
         * 
         * | Positional        | Regular         | VariadicPositional | Keyword        | VariadicKeyword  |
         * |-------------------|-----------------|--------------------|----------------|------------------|
         * | 0: [type1, type2] | [type3]         | ...                | [type5, type6] | [type7]          |
         * | 1: [type8]        |                 |                    | [type12]       |                  |
         * 
         * etc. 
         */
        // Skip all this malarky if there are no parameters at all
        if (Count is 0)
        {
            yield return new PythonFunctionParameterList<TResult>();
            yield break;
        }

        var options = Map(positionalProjector, regularProjector, variadicPositionalProjector, keywordProjector, variadicKeywordProjector);

        var ranges = options.Enumerable(x => x.Count(), x => x.Count(), x => x.Count(), x => x.Count(), x => x.Count());

        ImmutableArray<(ParameterGroup Group, IEnumerable<TResult> Choices)> ps = [
            ..options.Enumerable(p => (ParameterGroup.Positional, p),
                                 p => (ParameterGroup.Regular, p),
                                 p => (ParameterGroup.VariadicPositional, p),
                                 p => (ParameterGroup.Keyword, p),
                                 p => (ParameterGroup.VariadicKeyword, p))
        ];

        foreach (var turn in Permutations(ranges))
        {
            var selections = turn.Zip(ps, (i, p) => (p.Group, Value: p.Choices.ElementAt(i)))
                                 .ToLookup(e => e.Group, e => e.Value);

            yield return new PythonFunctionParameterList<TResult>(
                [.. selections[ParameterGroup.Positional]],
                [.. selections[ParameterGroup.Regular]],
                selections[ParameterGroup.VariadicPositional].FirstOrDefault(),
                [.. selections[ParameterGroup.Keyword]],
                selections[ParameterGroup.VariadicKeyword].FirstOrDefault());
        }
    }
```

The look-up at the end could be made another general helper.

---
Below is the diff if you want to apply with `git apply`:

```diff
diff --git a/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameter.cs b/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameter.cs
index 4b913348..37f2d66a 100644
--- a/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameter.cs
+++ b/src/CSnakes.SourceGeneration/Parser/Types/PythonFunctionParameter.cs
@@ -129,6 +129,15 @@ public sealed class PythonFunctionParameterList<T>(ImmutableArray<T> positional
         }
     }
 
+    enum ParameterGroup
+    {
+        Positional,
+        Regular,
+        VariadicPositional,
+        Keyword,
+        VariadicKeyword
+    }
+
     public IEnumerable<PythonFunctionParameterList<TResult>>
         MapMany<TResult>(Func<T, IEnumerable<TResult>> positionalProjector,
                          Func<T, IEnumerable<TResult>> regularProjector,
@@ -147,56 +156,38 @@ public sealed class PythonFunctionParameterList<T>(ImmutableArray<T> positional
          * etc. 
          */
         // Skip all this malarky if there are no parameters at all
-        if (Positional.IsEmpty && Regular.IsEmpty && VariadicPositional is not { } && Keyword.IsEmpty && VariadicKeyword is not { })
+        if (Count is 0)
         {
             yield return new PythonFunctionParameterList<TResult>();
             yield break;
         }
 
-        var positionalOptions = Positional.Select(positionalProjector);
-        var regularOptions = Regular.Select(regularProjector);
-        var variadicOption = VariadicPositional is { } vpp ? variadicPositionalProjector(vpp) : null;
-        var keywordOptions = Keyword.Select(keywordProjector);
-        var variadicKeywordOption = VariadicKeyword is { } vkp ? variadicKeywordProjector(vkp) : null;
+        var options = Map(positionalProjector, regularProjector, variadicPositionalProjector, keywordProjector, variadicKeywordProjector);
+
+        var ranges = options.Enumerable(x => x.Count(), x => x.Count(), x => x.Count(), x => x.Count(), x => x.Count());
 
-        List<int> ranges = [
-            .. positionalOptions.Select(x => x.Count()),
-            .. regularOptions.Select(x => x.Count()),
+        ImmutableArray<(ParameterGroup Group, IEnumerable<TResult> Choices)> ps = [
+            ..options.Enumerable(p => (ParameterGroup.Positional, p),
+                                 p => (ParameterGroup.Regular, p),
+                                 p => (ParameterGroup.VariadicPositional, p),
+                                 p => (ParameterGroup.Keyword, p),
+                                 p => (ParameterGroup.VariadicKeyword, p))
         ];
-        if (variadicOption is not null)
-            ranges.Add(variadicOption.Count());
-        ranges.AddRange(keywordOptions.Select(x => x.Count()));
-        if (variadicKeywordOption is not null)
-            ranges.Add(variadicKeywordOption.Count());
-
-        foreach (var turn in Permutations(ranges)) {
-            List<TResult> positionalTurn = new(positionalOptions.Count());
-            for (int i = 0; i < positionalOptions.Count(); i++)
-                positionalTurn.Add(positionalOptions.ElementAt(i).ElementAt(turn[i]));
-            List<TResult> regularTurn = new(regularOptions.Count());
-            for (int i = 0; i < regularOptions.Count(); i++)
-                regularTurn.Add(regularOptions.ElementAt(i).ElementAt(turn[Positional.Length + i]));
-            TResult? variadicPositionalTurn = null;
-            if (variadicOption is not null)
-                variadicPositionalTurn = variadicOption.ElementAt(turn[Positional.Length + Regular.Length]);
-            List<TResult> keywordTurn = new(keywordOptions.Count());
-            for (int i = 0; i < keywordOptions.Count(); i++)
-                keywordTurn.Add(keywordOptions.ElementAt(i).ElementAt(turn[Positional.Length + Regular.Length + (VariadicPositional is not null ? 1 : 0) + i]));
-            TResult? variadicKeywordTurn = null;
-            if (variadicKeywordOption is not null)
-                variadicKeywordTurn = variadicKeywordOption.ElementAt(turn[^1]);
+
+        foreach (var turn in Permutations(ranges))
+        {
+            var selections = turn.Zip(ps, (i, p) => (p.Group, Value: p.Choices.ElementAt(i)))
+                                 .ToLookup(e => e.Group, e => e.Value);
 
             yield return new PythonFunctionParameterList<TResult>(
-                [.. positionalTurn],
-                [.. regularTurn],
-                variadicPositionalTurn,
-                [.. keywordTurn],
-                variadicKeywordTurn
-                );
+                [.. selections[ParameterGroup.Positional]],
+                [.. selections[ParameterGroup.Regular]],
+                selections[ParameterGroup.VariadicPositional].FirstOrDefault(),
+                [.. selections[ParameterGroup.Keyword]],
+                selections[ParameterGroup.VariadicKeyword].FirstOrDefault());
         }
     }
 
-
     public int Count => Positional.Length
                 + Regular.Length
                 + (VariadicPositional is not null ? 1 : 0)
```

atifaziz (commented):
There's a lot to chew here, but started with an initial review and shared some comments.

src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs
```diff
@@ -13,62 +13,76 @@ public enum ConversionDirection
...
             ({ Name: "tuple" or "typing.Tuple" or "Tuple", Arguments: var ts }, _) => CreateTupleType(ts, direction),
             ({ Name: "dict" or "typing.Dict" or "Dict" or "typing.Mapping" or "Mapping", Arguments: [var kt, var vt] }, _) => CreateDictionaryType(kt, vt, direction),
-            ({ Name: "typing.Optional" or "Optional", Arguments: [var t] }, _) => SyntaxFactory.NullableType(AsPredefinedType(t, direction)),
+            ({ Name: "typing.Optional" or "Optional", Arguments: [var t] }, _) => AsPredefinedType(t, direction).Select(SyntaxFactory.NullableType),
             ({ Name: "typing.Generator" or "Generator", Arguments: [var yt, var st, var rt] }, _) => CreateGeneratorType(yt, st, rt, direction),
             ({ Name: "typing.Coroutine" or "Coroutine", Arguments: [var yt, var st, var rt] }, _) => CreateCoroutineType(yt, st, rt, direction),
+            ({ Name: "typing.Union" or "Union", Arguments: var ts }, ConversionDirection.ToPython) => [.. ts.SelectMany(t => AsPredefinedType(t, direction))],
+            ({ Name: "typing.AnyStr" or "AnyStr"}, ConversionDirection.ToPython) => [SyntaxFactory.ParseTypeName("string"), SyntaxFactory.ParseTypeName("byte[]")],
```
atifaziz, line 26: Is this the right interpretation of [`AnyStr`](https://docs.python.org/3/library/typing.html#typing.AnyStr), which is a _constrained generic type variable_ rather than a constructed/closed type? Here it seems to be treated like a `Union`. You can see that in the following test:

https://github.com/tonybaloney/CSnakes/blob/20b662e623719706f254f3435fca07a0fc06b2aa/src/Integration.Tests/python/test_overload.py#L17-L26

which now expands to this:

```c#
public interface ITestClass : IReloadableModuleImport
{
    // ...others omitted for brevity...

    /// <summary>
    /// Invokes the Python function <c>test_overload_unsupported_type</c>:
    /// <code><![CDATA[
    /// def test_overload_unsupported_type(x: AnyStr) -> str: ...
    /// ]]></code>
    /// </summary>
    string TestOverloadUnsupportedType(string x);

    /// <summary>
    /// Invokes the Python function <c>test_overload_unsupported_type</c>:
    /// <code><![CDATA[
    /// def test_overload_unsupported_type(x: AnyStr) -> str: ...
    /// ]]></code>
    /// </summary>
    string TestOverloadUnsupportedType(byte[] x);
}
```

This technically means that an unsupported case is now supported, yet it's still called unsupported? What's more, it's generating 3 overloads in the C# code!

src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_overload.approved.txt
```diff
@@ -173,6 +199,22 @@ public interface ITestClass : IReloadableModuleImport
...
+    /// </summary>
+    string TestOverloadUnsupportedType(string x);
+
+    /// <summary>
+    /// Invokes the Python function <c>test_overload_unsupported_type</c>:
+    /// <code><![CDATA[
+    /// def test_overload_unsupported_type(x: AnyStr) -> str: ...
+    /// ]]></code>
+    /// </summary>
+    string TestOverloadUnsupportedType(byte[] x);
```
atifaziz, line 217: These shouldn't get generated. See my [earlier comment](https://github.com/tonybaloney/CSnakes/pull/569/files#r2178089017).

src/CSnakes.SourceGeneration/Reflection/TypeReflection.cs
tonybaloney: Reading the docs it's also deprecated in 3.13 and since we don't support generics (yet) would be confusing. I'll remove it from this PR

src/CSnakes.Runtime.Tests/CSnakes.Runtime.Tests.csproj
```diff
@@ -4,6 +4,7 @@
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
     <NoWarn>$(NoWarn);PRTEXP001;PRTEXP002</NoWarn>
+    <Platforms>AnyCPU;x64</Platforms>
```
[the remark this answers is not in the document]
tonybaloney: CSnakes test suite doesn't work well on Windows on ARM. This is a way to force x64 emulation builds. Unrelated to this PR but I need it to test on anyway

src/CSnakes.SourceGeneration/Parser/EnumerableExtensions.cs
```diff
@@ -31,4 +31,37 @@ internal static partial class EnumerableExtensions
             }
         }
     }
+
+    public static IEnumerable<int[]> Permutations(this IEnumerable<int> ranges)
+    {
+        // Given a list of ranges, e.g.
+        // [2, 3, 1] (2 options for first, 3 for second, 1 for third), return all combinations of indexes
+        // e.g. [0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 1, 1], [1, 0, 0], [1, 0, 1], [1, 1, 0], [1, 1, 1]
```
atifaziz, line 39: Shouldn't this read as follows?

```suggestion
        // e.g. [0, 0, 0] [0, 1, 0] [0, 2, 0] [1, 0, 0] [1, 1, 0] [1, 2, 0]
```

tonybaloney:
@atifaziz I'm happy with this now. It's ready for final review

atifaziz (changes requested):
> @atifaziz I'm happy with this now. It's ready for final review

Can we bring in the fix from #499 and make the CI go green? Should be super easy to do. See [my earlier comment about this](https://github.com/tonybaloney/CSnakes/pull/569#discussion_r2180596503).

tonybaloney:
> > @atifaziz I'm happy with this now. It's ready for final review
> 
> 
> 
> Can we bring in the fix from #499 and make the CI go green? Should be super easy to do. See [my earlier comment about this](https://github.com/tonybaloney/CSnakes/pull/569#discussion_r2180596503).

I split the lines so there shouldn't be any test failures from that particular bug

atifaziz:
> I split the lines so there shouldn't be any test failures from that particular bug

Right, looks like then [a test is failing due to a new problem](https://github.com/tonybaloney/CSnakes/actions/runs/16078493534/job/45378838076?pr=569#step:11:41):

      Failed CSnakes.Tests.GeneratedSignatureTests.TestGeneratedSignature(code: "def t() -> str | int: \n  ...\n", expected: "PyObject T()") [47 ms]
      Error Message:
       Assert.Equal() Failure: Strings differ
                               ↓ (pos 16)
    Expected: "public PyObject T()"
    Actual:   "public PyObject t()"
                               ↑ (pos 16)
      Stack Trace:
         at CSnakes.Tests.GeneratedSignatureTests.TestGeneratedSignature(String code, String expected) in /_/src/CSnakes.Tests/GeneratedSignatureTests.cs:line 68
       at InvokeStub_GeneratedSignatureTests.TestGeneratedSignature(Object, Span`1)
       at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
    Test run for /home/runner/work/CSnakes/CSnakes/src/CSnakes.Runtime.Tests/bin/Debug/net9.0/CSnakes.Runtime.Tests.dll (.NETCoreApp,Version=v9.0)

tonybaloney:
I had a commit that didn't push from the plane :-) 

All passing again

PR #574, opened 2025-07-06
Add `dotnet new` template for console applications

atifaziz:
This PR adds a `dotnet new` template to get up and started with using CSnakes a breeze (helping with #314).

Once the NuGet package is published, users will be able to install via:

    dotnet new install CSnakes.Templates

and then create and run a new application using Python as simple as this:

    mkdir MyApp
    cd MyApp
    dotnet new pyapp
    dotnet run

See the updated documentation for more details. See also the read-me document in the `templates` directory for details on testing.

The template supports options to set the Python version, disable virtual environment support and select the package manager:

    Usage:
      dotnet new pyapp [options] [template options]
    
    Options:
      -n, --name <name>       The name for the output being created. If no name is specified, the name of the output directory is used.
      -o, --output <output>   Location to place the generated output.
      --dry-run               Displays a summary of what would happen if the given command line were run if it would result in a template creation.
      --force                 Forces content to be generated even if it would change existing files.
      --no-update-check       Disables checking for the template package updates when instantiating a template.
      --project <project>     The project that should be used for context evaluation.
      -lang, --language <C#>  Specifies the template language to instantiate.
      --type <project>        Specifies the template type to instantiate.
    
    Template options:
      -P, --PythonVersion <3.10|3.11|3.12|3.13|3.9>  Python version to target.
                                                     Type: choice
                                                       3.9   Target Python 3.9
                                                       3.10  Target Python 3.10
                                                       3.11  Target Python 3.11
                                                       3.12  Target Python 3.12
                                                       3.13  Target Python 3.13
                                                     Default: 3.12
      -N, --NoVirtualEnvironment                     Use a virtual environment for Python dependencies.
                                                     Type: bool
                                                     Default: false
      -Pa, --PackageManager <none|pip|uv>            Package manager to use.
                                                     Type: choice
                                                       none  No package manager.
                                                       pip   Use pip as the package manager.
                                                       uv    Use uv as the package manager.
                                                     Default: none

### DIFF
--- a/templates/src/ConsoleApp/.template.config/template.json
+++ b/templates/src/ConsoleApp/.template.config/template.json
@@ -5,7 +5,8 @@
     "Python",
     "Console"
   ],
-  "identity": "CSnakes.Templates.Console",
+  "groupIdentity": "CSnakes.Templates.Console",
+  "identity": "CSnakes.Templates.Console.CSharp",
   "name": "Console App (CSnakes)",
   "shortName": "pyapp",
   "tags": {


templates/tests/pyapp.--NoVirtualEnvironment#--PackageManager#pip.verified/pyapp/pyapp.csproj
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include=".template.config\template.json" />
+    <AdditionalFiles Include="*.py">
```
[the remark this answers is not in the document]
atifaziz: Not quite. The [template is bound to version 1.1](https://github.com/tonybaloney/CSnakes/pull/574/files#diff-9e6c79b04bdfb202aa42c7af52b93e31c435727eb55317b7d3ba133e2bebc09cR18) so first 1.2 will need to be released, then the template updated to refer to 1.2 and finally the template package published.

---
PS You might also need new workflows for templates, which this PR doesn't address.

templates/src/ConsoleApp/.template.config/template.json
```diff
@@ -0,0 +1,119 @@
+{
+  "$schema": "http://json.schemastore.org/template",
+  "author": "Atif Aziz",
+  "classifications": [
+    "Python",
+    "Console"
+  ],
+  "identity": "CSnakes.Templates.Console",
```
sayedihashimi, line 8: If you ever plan to release another template for a different language you may want to change this to `CSnakes.Templates.Console.CSharp` so that later you can add `groupIdentity: CSnakes.Templates.Console`. The benefit of this is that you can have a single template that appears as `CSnakes.Tempates.Console` but supports multiple languages and those templates have their own template.config and content. If this will never happen, then no change needed for this field.

templates/src/ConsoleApp/.template.config/template.json
```diff
@@ -0,0 +1,119 @@
...
+    },
+    "UsePipInstaller": {
+      "type": "computed",
+      "value": "PackageManager == \"pip\""
+    },
+    "UseUvInstaller": {
+      "type": "computed",
+      "value": "PackageManager == \"uv\""
+    }
+  }
```
sayedihashimi, line 118: You may want to add `primaryOutputs` which should point to the generated project file. Here is a reference https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/RazorClassLibrary-CSharp/.template.config/template.json#L75.

templates/src/ConsoleApp/ConsoleApp.csproj
```diff
@@ -0,0 +1,23 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include=".template.config\template.json" />
```
sayedihashimi: Is it intended that this is in the csproj file? After the template is used this would be in the users generated project file, unless I've misunderstood something about this template.

templates/src/ConsoleApp/.template.config/template.json
```diff
@@ -0,0 +1,119 @@
...
+    "Console"
+  ],
+  "identity": "CSnakes.Templates.Console",
+  "name": "Console App (CSnakes)",
+  "shortName": "pyapp",
+  "tags": {
+    "language": "C#",
+    "type": "project"
+  },
+  "sourceName": "ConsoleApp",
```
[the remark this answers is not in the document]
sayedihashimi: Looks good

templates/src/ConsoleApp/.template.config/template.json
atifaziz: I used the [C# console template](https://github.com/dotnet/sdk/blob/795b68c1279ce6d6b53d80cfcec6673d9f660568/template_feed/Microsoft.DotNet.Common.ProjectTemplates.10.0/content/ConsoleApplication-CSharp/.template.config/template.json#L134-L142) for inspiration and didn't take over this part because of the condition the same condition there (`(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")`). I wanted to address the `dotnet` CLI with this first take and add IDE/host support if and when desired.

templates/src/ConsoleApp/.template.config/template.json
atifaziz: We've been primarily focusing on C# with source generators, but I guess there's no harm in keeping the doors open here. After all, [we do have a sample in F# too](https://github.com/tonybaloney/CSnakes/tree/1a75be9797b807de0e8372af6ddf43a9c1efb67d/samples/simple/FSharpSample).

Added group identity with be8378edf366f87add0579db2461a0e1d2244db8.

PR #576, opened 2025-07-07
Integrate source generator more tightly with build system

atifaziz:
This PR makes managing `<AdditionalFiles>` element optional for CSnakes (helping with #314). It adds a new build targets file to the NuGet package that automatically recognises all `*.py` files as _additional files_. This new default behaviour requires less setup and ceremony for the user. Should someone desire more control, it can be disabled by setting the new `DisableDefaultPythonItems` property to `true` in the project file. A new build properties file has also been added to the NuGet package that sets it to `false`. This file isn't strictly required but is being added for convenience. When `DisableDefaultPythonItems` is set to `true`, the user must manually add the `<AdditionalFiles>` element like before though it requires a new property/attribute. See the updated documentation for more.

Yet another new property being recognised by the source generator is `EmbedPythonSources`, which isn't defined by default. When set to `true`, however, the Python source files are embedded into the generated code. This replaces the `csnakes_embed_source` option that needed a `.globalconfig` file:

https://github.com/tonybaloney/CSnakes/blob/20b662e623719706f254f3435fca07a0fc06b2aa/src/RedistributablePython.Tests/.globalconfig#L1

Managing everything in the project file seems simpler, more conventional and even compatible with [project-less C# apps landing with .NET 10](https://devblogs.microsoft.com/dotnet/announcing-dotnet-run-app/#setting-msbuild-properties-with-#:property). This property still remains _undocumented_. If we want to make this more obvious then it could be renamed to `_EmbedPythonSources`.


tonybaloney:
@atifaziz I updated the docs in the PR and renamed the setting, can you change it to be off by default pls

atifaziz:
> @atifaziz I updated the docs in the PR and renamed the setting, can you change it to be off by default pls

@tonybaloney Done with a658f9e4c0961dfa060c41347c60f332e412dcf5 and f677bbaac46adb67bd0b973bdb4561192892e3e3.

Since [this is approved](https://github.com/tonybaloney/CSnakes/pull/576#pullrequestreview-3050739990), I am going to go ahead and merge it since the [CI is green on the latest commit](https://github.com/tonybaloney/CSnakes/actions/runs/16848980695). However, do take note that `SourceItemType="Python"` is now required on `AdditionalFiles`:

```xml
<ItemGroup>
    <AdditionalFiles Include="*.py" SourceItemType="Python">
    </AdditionalFiles>
</ItemGroup>
```

If this is distasteful, we can debate and change with a future PR. Meanwhile, it requires explicit targeting, which avoids accidents and affords some flexibility (the source generator no longer cares of about the file extension and process additional files with this property). When `DefaultPythonItems` is used, `SourceItemType="Python"` is set automatically.

Issue #582, opened 2025-07-18
Environment Execute Globals Not Updating


rc14193:
```using CSnakes.Runtime;
using TestCSnakes.Components;

var builder = WebApplication.CreateBuilder(args);

// Add services to the container.
builder.Services.AddRazorComponents()
    .AddInteractiveServerComponents();
builder.Services
    .WithPython()
    .WithHome(Environment.CurrentDirectory)
    .FromRedistributable();

var app = builder.Build();

// Configure the HTTP request pipeline.
if (!app.Environment.IsDevelopment())
{
    app.UseExceptionHandler("/Error", createScopeForErrors: true);
    // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
    app.UseHsts();
}

app.UseHttpsRedirection();


app.UseAntiforgery();

app.MapStaticAssets();
app.MapRazorComponents<App>()
    .AddInteractiveServerRenderMode();

var env = app.Services.GetRequiredService<IPythonEnvironment>();
string code = @"
x['test'] = '5'
print(t)
t = 'mine'
print(t)
n = 7
";
var x = CSnakes.Runtime.Python.PyObject.From(new Dictionary<string, string>() { ["z"] = "eeee" });
var globals = new Dictionary<string, CSnakes.Runtime.Python.PyObject>() { ["x"] = x, ["t"] = CSnakes.Runtime.Python.PyObject.From("jest") };
var locals = new Dictionary<string, CSnakes.Runtime.Python.PyObject>() { };
using var res = env.Execute(code, locals, globals);

app.Run();
```

<img width="278" height="187" alt="Image" src="https://github.com/user-attachments/assets/2e25767e-56ec-4051-8c7e-0b7ad828a9a7" />

<img width="477" height="145" alt="Image" src="https://github.com/user-attachments/assets/ee85898c-75bc-4ab7-af5a-8cacc445be63" />

Minimum example above in a new net 9.0 blazor app with visual studio. I expect the dictionary to contain a key z and test. It does. I expect t to update from jest to mine. It does not. Is this the expected behavior?

tonybaloney:
Shipping in next update

PR #600, opened 2025-07-31
Align Python Exec

rc14193:
Addresses #582 by converting the marshalled dictionaries used in Py_RunString back to the C# Dictionary<string, PyObject> and explicitly updating the globals and locals passed to the CSnakes functions.


rc14193:
I apologize for the closed and now duplicate. I wanted to correct my username on the commit.

tonybaloney:
Thanks, I added a test and refactored the code slightly to avoid allocating a Dictionary instance to merge the original dictionary where it isn't required.

tonybaloney (approved):

PR #687, opened 2025-09-16
Allow conversions from Nullable<T> to T? and catch cases when V = None

tonybaloney:


src/CSnakes.Runtime/Python/PyObject.cs
```diff
@@ -658,8 +659,10 @@ internal object As(Type type)
                 var t when t.IsAssignableTo(typeof(ITuple)) => PyObjectTypeConverter.ConvertToTuple(this, t),
                 var t when t.IsAssignableTo(typeof(IGeneratorIterator)) => PyObjectTypeConverter.ConvertToGeneratorIterator(this, t),
                 var t when t.IsAssignableTo(typeof(ICoroutine)) => PyObjectTypeConverter.ConvertToCoroutine(this, t),
+                var _ when Nullable.GetUnderlyingType(type) != null => IsNone() ? null : As(Nullable.GetUnderlyingType(type)!),
```
atifaziz, line 662: Remove duplicate call to `Nullable.GetUnderlyingType`:

```suggestion
                var t when Nullable.GetUnderlyingType(t) is { } vt => IsNone() ? null : As(vt)
```

src/CSnakes.Tests/PythonStaticGeneratorTests/FormatClassFromMethods.test_optional.approved.txt
```diff
@@ -133,6 +138,19 @@ public static class TestClassExtensions
...
+
+        public (long? , string) TestOptionalTuple((long? , string) a)
+        {
+            using (GIL.Acquire())
+            {
+                this.logger?.LogDebug("Invoking Python function: {FunctionName}", "test_optional_tuple");
+                PyObject __underlyingPythonFunc = this.__func_test_optional_tuple;
+                using PyObject a_pyObject = PyObject.From(a)!;
+                using PyObject __result_pyObject = __underlyingPythonFunc.Call(a_pyObject);
+                var __return = __result_pyObject.BareImportAs<(long? , string), global::CSnakes.Runtime.Python.PyObjectImporters.Tuple<long?, string, global::CSnakes.Runtime.Python.PyObjectImporters.Runtime<long?>, global::CSnakes.Runtime.Python.PyObjectImporters.String>>();
```
atifaziz, line 150: Unfortunately, `global::CSnakes.Runtime.Python.PyObjectImporters.Runtime<long?>` here means that `long?` is imported via run-time reflection whereas normally its importer would be resolved at build-time. This introduces subtle differences in when Native AOT would be supported.

PR #692, opened 2025-09-17
🚧 Add dynamic code attributes for flagging Native AOT compatibility

### LINKED ISSUE #679: Investigate remaining IL20** and IL3050 warnings and their implication on NAOT scenarios
There are several warnings in the build for the original reflection code that are incompatible with NAOT. Since we emit trimming on the assembly, they need to be resolved.

1. Can we refactor the code so that it doesn't require dynamic attribute access?
2. What code paths does this impact?
3. Does this impact source generated code?

atifaziz:
This PR enhances CSnakes' compatibility with Native AOT scenarios by adding appropriate `[RequiresDynamicCode]` and `[DynamicallyAccessedMembers]` attributes throughout the type conversion system. The changes ensure that code paths requiring dynamic type creation are properly annotated, helping developers understand AOT limitations and enabling better tooling support.

All ILXXXX warnings should disappear with this PR. It effectively closes #679.

The core idea is to clearly demarcate code that requires dynamic capabilities from code that can work in AOT scenarios. This is achieved through:

1. **Explicit Dynamic Code Annotations**: Adding `[RequiresDynamicCode]` attributes to methods that use `Type.MakeGenericType()` or similar dynamic operations
2. **Member Access Annotations**: Using `[DynamicallyAccessedMembers]` to indicate which type members need to be preserved for reflection
3. **Importer Stratification**: Introducing specialized importers that avoid dynamic code where possible

### Type Converter Annotations

The `PyObjectTypeConverter` class and its partial implementations now carry `[RequiresDynamicCode]` attributes on methods that perform dynamic type construction:

- `ConvertToCoroutine()` - Creates generic coroutine types dynamically
- `ConvertToDictionary()` - Constructs generic dictionary types
- `ConvertToGeneratorIterator()` - Builds generic generator iterator types
- `ConvertToList()` - Creates generic list types
- `ConvertToTuple()` - Constructs tuple types with dynamic member access

### PyObject API Enhancements

The `PyObject.As<T>()` methods are now properly annotated with `[RequiresDynamicCode]` since they delegate to the dynamic type conversion system. This provides clear guidance to developers about which APIs require dynamic capabilities.

#### New `None` Importer

A new `PyObjectImporters.None` class has been introduced to handle `None` values without requiring dynamic code generation. This provides a AOT-compatible path for handling Python's `None` type.

This change has been forked into PR #690 and can be merged independently, thus reducing the delta here.

#### Runtime Importer Documentation

The `PyObjectImporters.Runtime<T>` class now includes comprehensive documentation explaining its dynamic code requirements and recommending careful review when used in AOT scenarios.

### Generated Code Optimizations

The source generator now produces more AOT-friendly code by:

- Using `PyObjectImporters.None` instead of `PyObjectImporters.Runtime<PyObject>` for return types that should be `None`
- Using `PyObjectImporters.Clone` instead of `PyObjectImporters.Runtime<PyObject>` for generic `PyObject` handling

These changes has been forked into PR #690 and can be merged independently, thus reducing the delta here.

### String Execution Optimizations

The `PythonRunString` class has been optimized to avoid unnecessary dynamic type conversions by introducing a specialized `Merge()` overload that uses statically-typed importers instead of the runtime conversion system.

This PR introduces new attributes and type annotations but maintains full backward compatibility. The only visible changes are:

- Additional compiler warnings in AOT scenarios (which is the intended behavior)
- Updated generated code that uses more specific importers (but maintains the same runtime behavior)

### DIFF
--- a/src/CSnakes.Runtime/PyObjectTypeConverter.cs
+++ b/src/CSnakes.Runtime/PyObjectTypeConverter.cs
@@ -1,6 +1,7 @@
 using CSnakes.Runtime.CPython;
 using CSnakes.Runtime.Python;
 using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 
 namespace CSnakes.Runtime;
@@ -8,7 +9,10 @@ internal partial class PyObjectTypeConverter
 {
     private static readonly ConcurrentDictionary<Type, DynamicTypeInfo> knownDynamicTypes = [];
 
-    public static object PyObjectToManagedType(PyObject pyObject, Type destinationType)
+    [RequiresDynamicCode("Calls System.Type.MakeGenericType(params Type[])")]
+    public static object PyObjectToManagedType(PyObject pyObject,
+                                               [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)]
+                                               Type destinationType)
     {
         if (CPythonAPI.IsPyDict(pyObject) && IsAssignableToGenericType(destinationType, dictionaryType))
         {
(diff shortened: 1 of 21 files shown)


tonybaloney:
+1 in principle after rebasing

src/CSnakes.Runtime/PyObjectTypeConverter.cs
```diff
@@ -31,11 +35,6 @@ public static object PyObjectToManagedType(PyObject pyObject, Type destinationTy
             return ConvertToList(pyObject, destinationType);
         }
 
-        if (CPythonAPI.IsBuffer(pyObject) && destinationType.IsAssignableTo(bufferType))
```
atifaziz, line 34: @tonybaloney I've moved this to `PyObject.As(Type type)`. See [my other comment for why](https://github.com/tonybaloney/CSnakes/pull/692/files#r2358004640).

tonybaloney (approved):

PR #697, opened 2025-09-18
Log & test logging queue buffer overflow

atifaziz:
This PR enhances the Python logging bridge to track and report dropped messages instead of silently losing them. Previously, when the logging queue overflowed, messages would be silently lost. Now the system provides clear indication of how many messages were dropped, enabling better monitoring and debugging.

The generator now yields a 4-tuple instead of a 3-tuple:

```python
# Before: (level, message, exception_info)
# After:  (drop_count, level, message, exception_info)
```

The drop count represents messages lost since the last successful yield, and is reset to zero after being reported.

The bridge now processes drop counts and issues debug messages when drops are detected.

`test_many_entries(count: int)` replaces the hardcoded `test_fifty_entries()` to allow configurable test scenarios.

**Overflow Test**: A comprehensive test that deliberately creates a slow logger scenario to verify drop count functionality:

- Uses barriers and countdown events for precise synchronization
- Blocks the first log message to fill the buffer
- Verifies that exactly 200 messages (buffer size) are preserved
- Confirms older messages are properly dropped

**Performance Impact**: The overhead of drop counting is fairly minimal:

- An atomic increment protected by a lock that's only acquired during overflow conditions.
- An additional `long` is marshaled as a tuple element.

### DIFF
--- a/src/CSnakes.Runtime/PythonLogger.cs
+++ b/src/CSnakes.Runtime/PythonLogger.cs
@@ -147,7 +147,7 @@ file sealed class Bridge(PyObject handler, PyObject uninstallCSnakesHandler, Tas
                     // this loop isn't keeping up due to the logger being slow or blocking. Adding
                     // to the misery wouldn't be helpful so issue a debug message instead.
 
-                    Debug.WriteLine($"Dropped {dropCount} log messages due to buffer overflow.");
+                    Debug.WriteLine($"Dropped {dropCount} log messages due to log buffer overflow.");
                 }
 
                 try
--- a/src/CSnakes.Runtime/PythonLogger.cs
+++ b/src/CSnakes.Runtime/PythonLogger.cs
@@ -23,12 +23,15 @@ file sealed class Bridge(PyObject handler, PyObject uninstallCSnakesHandler, Tas
     const string ModuleCode = """
         import logging
         import queue
+        import threading
 
 
         class __csnakesMemoryHandler(logging.Handler):
             def __init__(self):
                 logging.Handler.__init__(self)
                 self.queue = queue.Queue(200)
+                self.stats_lock = threading.Lock()
+                self.drop_count = 0
 
             def emit(self, record):
                 _ = self._put(record)
@@ -42,6 +45,8 @@ file sealed class Bridge(PyObject handler, PyObject uninstallCSnakesHandler, Tas
                         try:
                             _ = self.queue.get_nowait()  # drop the oldest record
                             self.queue.task_done()
+                            with self.stats_lock:
+                                self.drop_count += 1
                         except queue.Empty:
                             pass
                 return False  # all attempts to put failed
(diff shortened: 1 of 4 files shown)


src/CSnakes.Runtime/PythonLogger.cs
```diff
@@ -126,7 +135,20 @@ private static Task StartRecordListener(PyObject handler, ILogger logger)
...
+                {
+                    // One could log a warning here:
+                    //
+                    //   logger.LogWarning("Dropped {DropCount} log messages due to buffer overflow.", dropCount);
+                    //
+                    // but the reason the messages are getting dropped in the first place is because
+                    // this loop isn't keeping up due to the logger being slow or blocking. Adding
+                    // to the misery wouldn't be helpful so issue a debug message instead.
+
+                    Debug.WriteLine($"Dropped {dropCount} log messages due to buffer overflow.");
```
tonybaloney, line 150:
```suggestion
                    Debug.WriteLine($"Dropped {dropCount} log messages due to log buffer overflow.");
```

tonybaloney: Only because "buffer overflow" sounds scary in the wrong context

src/CSnakes.Runtime/PythonLogger.cs
```diff
@@ -109,10 +117,11 @@ private static Task StartRecordListener(PyObject handler, ILogger logger)
     {
         using PyObject getRecords = handler.GetAttr("get_records");
         using PyObject getRecordsResult = getRecords.Call();
-        IGeneratorIterator<(long, string, ExceptionInfo?), PyObject, PyObject> generator =
-            getRecordsResult.ImportAs<IGeneratorIterator<(long, string, ExceptionInfo?), PyObject, PyObject>,
-                                                         PyObjectImporters.Generator<(long, string, ExceptionInfo?), PyObject, PyObject,
-                                                                                     PyObjectImporters.Tuple<long, string, ExceptionInfo?,
+        IGeneratorIterator<(long, long, string, ExceptionInfo?), PyObject, PyObject> generator =
+            getRecordsResult.ImportAs<IGeneratorIterator<(long, long, string, ExceptionInfo?), PyObject, PyObject>,
+                                                         PyObjectImporters.Generator<(long, long, string, ExceptionInfo?), PyObject, PyObject,
```
tonybaloney, line 122: Is this change from the other importer PRs?

atifaziz: No, this is just adding extending the tuple with an additional `long` for the drop count, going from `(long, string, ExceptionInfo?)` to `(long, long, string, ExceptionInfo?)` Unfortunately, given the number of importers and their type arguments involved (exacerbated by the optional `ExceptionInfo` tuple being nested in another tuple), I had to keep things aligned with formatting/indentation for sanity's sake and so the delta here appears bigger than it is. The good news is that the compiler statically guarantees that this setup is correct otherwise it'll reject the compilation.

Issue #684, opened 2025-09-15
Add directive comment to skip functions


atifaziz:
Consider supporting a directive comment, like [`# type: ignore` comments](https://typing.python.org/en/latest/spec/directives.html#type-ignore-comments), that will cause the source generator to skip emitting a C# proxy for the Python function with the directive comment.

My proposal would be for the directive to be `# csharp: ignore` (alternatively `# csnakes: ignore`) and _must_ be applied to the same line where the function definition begins, as in:

```python
def foo():  # csharp: ignore
    pass
```

or:

```python
def foo(  # csharp: ignore
        bar,
        baz,
    ):
    pass
```

The following will not have the desired effect:

```python
def foo(
        bar,
        baz,
    ):  # csharp: ignore
    pass
```

The source generator should be prepared to handle the _directive comment_ being embedded within a comment such that both of the following work:

```python
def foo():  # type: ignore # csharp: ignore
    pass

def bar ():  # csharp: ignore # type: ignore
    pass
```

PR #427 already added support for skipping functions whose name starts with an underscore, but that doesn't prevent CSnake's Python parser from parsing the function. In the case of the directive comment proposed here, the parser should completely skip the function without any parsing. This could be as simple as extending the following condition (though that's not the suggested implementation):

https://github.com/tonybaloney/CSnakes/blob/8af6057218a26a83cafbd1332e25731fb41ce04f/src/CSnakes.SourceGeneration/Parser/PythonParser.Function.cs#L32-L33

This would help in two scenarios:

1. Developer doesn't wish to apply the underscore to the Python function
2. Prevent warnings when the source generator doesn't support a particular signature/syntax

An example of the second would be:

```python
P = ParamSpec("P")
T_co = TypeVar("T_co", covariant=True)


def my_decorator(func: Callable[P, T_co]) -> Callable[P, T_co]:  # csharp: ignore
    ...
```

maazarif11:
i can take this

PR #702, opened 2025-09-18
Adding ignore function directive

maazarif11:
A python function can have "# csharp: ignore" in-line to a `def func`, this will make the parser ignore the function.
https://github.com/tonybaloney/CSnakes/issues/684

### DIFF
--- a/src/CSnakes.Tests/GeneratedSignatureTests.cs
+++ b/src/CSnakes.Tests/GeneratedSignatureTests.cs
@@ -63,6 +63,9 @@ public class GeneratedSignatureTests
     [InlineData("def hello(a: str, b: int = 4, *, kw: str) -> None:\n ...\n", "void Hello(string a, string kw, long b = 4)")]
     [InlineData("def hello() -> str | int: \n  ...\n", "PyObject Hello()")]
     [InlineData("def escape(s: str, quote: bool = True) -> str: ...\n", "string Escape(string s, bool quote = true)")]
+    [InlineData("# csharp: ignore\ndef foo(bar):\n    pass", "PyObject Foo(PyObject bar)")]  // Comment on line above
+    [InlineData("def foo(\n    bar  # csharp: ignore\n):\n    pass", "PyObject Foo(PyObject bar)")]  // Comment in parameter list
+    [InlineData("def foo(bar):\n    # csharp: ignore\n    pass", "PyObject Foo(PyObject bar)")]  // Comment inside function body
     public void TestGeneratedSignature(string code, string expected)
     {
         SourceText sourceText = SourceText.From(code);
@@ -225,5 +228,18 @@ public class GeneratedSignatureTests
         Assert.Empty(errors1);
         var module1 = ModuleReflection.MethodsFromFunctionDefinitions(functions1).ToImmutableArray();
         Assert.Single(module1);
+    }
+
+    [Theory]
+    [InlineData("def foo(  # csharp: ignore\r\n        bar,\r\n        baz,\r\n    ):\r\n    pass\r\n")]
+    [InlineData("def hello(): # csharp: ignore ...\n")]
+    [InlineData("def foo():  # type: ignore # csharp: ignore\r\n    pass\r\n\r\ndef bar ():  # csharp: ignore # type: ignore\r\n    pass")]
+    public void TestCSharpIgnoreCommentFunction(string code)
+    {
+        SourceText sourceText = SourceText.From(code);
+        Assert.True(PythonParser.TryParseFunctionDefinitions(sourceText, out var functions, out var errors));
+        Assert.Empty(errors);
+        var module = ModuleReflection.MethodsFromFunctionDefinitions(functions).ToImmutableArray();
+        Assert.Empty(module);
     }
 }
--- a/src/CSnakes.SourceGeneration/Parser/PythonParser.Function.cs
+++ b/src/CSnakes.SourceGeneration/Parser/PythonParser.Function.cs
@@ -32,6 +32,14 @@ public static partial class PythonParser
     static bool IsFunctionSignature(string line) =>
         line.StartsWith("def ") || line.StartsWith("async def");
 
+    /// <summary>
+    /// Checks if the line contains a "# csharp: ignore" comment.
+    /// </summary>
+    /// <param name="line">Line to check</param>
+    /// <returns></returns>
+    static bool HasCSharpIgnoreComment(string line) =>
+        line.Contains("# csharp: ignore");
+
     public static bool TryParseFunctionDefinitions(SourceText source, out PythonFunctionDefinition[] pythonSignatures, out GeneratorError[] errors)
     {
         // Go line by line
@@ -45,6 +53,14 @@ public static partial class PythonParser
             string lineOfCode = line.ToString();
             if (!IsFunctionSignature(lineOfCode) && !unfinishedFunctionSpec)
             {
+                continue;
+            }
+
+            // Check for "# csharp: ignore" comment on the first line of function definition
+            if (IsFunctionSignature(lineOfCode) && HasCSharpIgnoreComment(lineOfCode))
+            {
+                currentBuffer = [];
+                unfinishedFunctionSpec = false;
                 continue;
             }
 


src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +229,17 @@ public void TestUnionReturnType(string returnType)
         var module1 = ModuleReflection.MethodsFromFunctionDefinitions(functions1).ToImmutableArray();
         Assert.Single(module1);
     }
+
+    [Theory]
+    [InlineData("def foo(  # csharp: ignore\r\n        bar,\r\n        baz,\r\n    ):\r\n    pass\r\n")]
+    [InlineData("def hello(): # csharp: ignore ...\n")]
+    [InlineData("def foo():  # type: ignore # csharp: ignore\r\n    pass\r\n\r\ndef bar ():  # csharp: ignore # type: ignore\r\n    pass")]
```
atifaziz, line 236: What about separating this into two cases so if one fails, it's easier to identify which?

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +229,17 @@ public void TestUnionReturnType(string returnType)
...
+    public void TestCSharpIgnoreCommentFunction(string code)
+    {
+        SourceText sourceText = SourceText.From(code);
+        Assert.True(PythonParser.TryParseFunctionDefinitions(sourceText, out var functions, out var errors));
+        Assert.Empty(errors);
+        var module = ModuleReflection.MethodsFromFunctionDefinitions(functions).ToImmutableArray();
+        Assert.Empty(module);
```
atifaziz, line 243: Wouldn't this be sufficient?

```suggestion
        Assert.Empty(functions);
```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +252,43 @@ public void TestUnionReturnType(string returnType)
...
+
+    [Theory]
+    [InlineData("""
+        
+                def foo(  # csharp: ignore
+                        bar,
+                        baz,
+                    ):
+                    pass
+
```
atifaziz, line 264: This indentation is incorrect. It produces the following string _literally_:

```

        def foo(  # csharp: ignore
                bar,
                baz,
            ):
            pass

```

In other words, the function is _indented_ and indentation is significant in Python.

This indentation happens because the closing `"""` decided the level indentation applied or where the common leading space is removed. The following rules govern the interpretation of a multi-line [raw string literal](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/raw-string):

> - The opening quotes must be the last non-whitespace characters on their line, and the closing quotes must be the first non-whitespace characters on their line.
> - **Any whitespace to the left of the closing quotes is removed from all lines of the raw string literal.**
> - Any whitespace following the opening quotes on the same line is ignored.
> - Whitespace only lines following the opening quote are included in the string literal.
> - If a whitespace precedes the end delimiter on the same line, the exact number and kind of whitespace characters (for example, spaces vs. tabs) must exist at the beginning of each content line. Specifically, a space does not match a horizontal tab, and vice versa.
> - **The newline before the closing quotes isn't included in the literal string.**

I've highlighted two to be mindful of. What you want is the following:

```suggestion
        def foo(  # csharp: ignore
                bar,
                baz,
            ):
            pass

```

The funny thing is that the test is incidentally passing because the function is being skipped due to `IsFunctionSignature` returning `false`. It never gets to the `HasCSharpIgnoreComment` check since it comes later. I think to make the tests more robust, it may wise to assert there is no common indentation across the code with something like this (although it's uncommon to assert on test inputs):

```c#
var indentations =
    from line in sourceText.Lines
    select line.ToString()
    into line
    where line.Length > 0
    select line.TakeWhile(ch => ch == ' ').Count();
Assert.Equal(0, indentations.Min());
```

---
PS The blank line before the closing `"""` is significant due to the last rule.

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +252,43 @@ public void TestUnionReturnType(string returnType)
...
+                        baz,
+                    ):
+                    pass
+        
+        """,
+        Label = "C# ignore comment - Multi lined function definition")]
+    [InlineData("""
+        
+                def hello(): # csharp: ignore ...
+
```
atifaziz, line 270: Fix indentation:

```suggestion
        def hello(): # csharp: ignore ...
            pass

```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +252,43 @@ public void TestUnionReturnType(string returnType)
...
+        
+                def hello(): # csharp: ignore ...
+        
+        """,
+        Label = "C# ignore comment - Single line function definition")]
+    [InlineData("""
+        
+                def foo():  # type: ignore # csharp: ignore
+                    pass
+
```
atifaziz, line 277: Fix indentation:

```suggestion
        def foo():  # type: ignore # csharp: ignore
            pass

```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +252,43 @@ public void TestUnionReturnType(string returnType)
...
+                def foo():  # type: ignore # csharp: ignore
+                    pass
+        
+        """,
+        Label = "C# ignore comment - Embedded within a comment after")]
+    [InlineData("""
+        
+                def bar():  # csharp: ignore # type: ignore
+                    pass
+
```
atifaziz, line 284: Fix indentation:

```suggestion
        def bar():  # csharp: ignore # type: ignore
            pass

```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -226,4 +252,43 @@ public void TestUnionReturnType(string returnType)
...
+        
+        """,
+        Label = "C# ignore comment - Embedded within a comment before")]
+    public void TestCSharpIgnoreCommentFunction(string code)
+    {
+        SourceText sourceText = SourceText.From(code);
+        Assert.True(PythonParser.TryParseFunctionDefinitions(sourceText, out var functions, out var errors));
+        Assert.Empty(errors);
+        Assert.Empty(functions);
+    }
```
atifaziz, line 293: Considering asserting that the input had no common indentation (perhaps with a comment as to why):

```suggestion
        var indentations =
            from line in sourceText.Lines
            select line.ToString()
            into line
            where line.Length > 0
            select line.TakeWhile(ch => ch == ' ').Count();
        Assert.Equal(0, indentations.Min());
    }
```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -63,6 +63,32 @@ public class GeneratedSignatureTests
     [InlineData("def hello(a: str, b: int = 4, *, kw: str) -> None:\n ...\n", "void Hello(string a, string kw, long b = 4)")]
     [InlineData("def hello() -> str | int: \n  ...\n", "PyObject Hello()")]
     [InlineData("def escape(s: str, quote: bool = True) -> str: ...\n", "string Escape(string s, bool quote = true)")]
+    [InlineData("""
+        
+        # csharp: ignore
+        def foo(bar):
+            pass
+        
+        """, "PyObject Foo(PyObject bar)",
```
atifaziz, line 72: nit: for consistency:

```suggestion
        """,
        "PyObject Foo(PyObject bar)",
```

src/CSnakes.Tests/GeneratedSignatureTests.cs
```diff
@@ -63,6 +63,32 @@ public class GeneratedSignatureTests
...
+        """,
+        "PyObject Foo(PyObject bar)",
+        Label = "C# ignore comment in parameter list")]
+    [InlineData("""
+        
+        def foo(bar):
+            # csharp: ignore
+            pass
+        
+        """, "PyObject Foo(PyObject bar)",
```
atifaziz, line 90: nit: for consistency:

```suggestion
        """,
        "PyObject Foo(PyObject bar)",
```

PR #724, opened 2025-10-07
Set ignore for XML comments on public

buvinghausen:


buvinghausen:
@atifaziz changes made

PR #778, opened 2025-12-30
Ensure completed task cancellation does not break event loop

### LINKED ISSUE #776: Accessing disposed object when sharing `CancellationToken`
Calling tasks sequentially while sharing a `CancellationToken` can make the `EventLoop` crash in .NET.

The `EventLoop` does [not unregister](https://github.com/tonybaloney/CSnakes/blob/022c069ff80e9d921aa2f40c4ad3bf639afc4977/src/CSnakes.Runtime/CPython/EventLoop.cs#L269-L270) the scheduling of cancellations of tasks that have already completed. Since we [dispose the underlying `PyObject` on conclusion](https://github.com/tonybaloney/CSnakes/blob/022c069ff80e9d921aa2f40c4ad3bf639afc4977/src/CSnakes.Runtime/CPython/EventLoop.cs#L95), subsequent cancellations will cause a crash in the event loop as we are trying to access an already disposed object:

```
System.ObjectDisposedException
  HResult=0x80131622
  Message=Cannot access a disposed object.
Object name: 'CSnakes.Runtime.Python.PyObject'.
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowObjectDisposedException(Object instance)
   at System.Runtime.InteropServices.SafeHandle.DangerousAddRef(Boolean& success)
   at CSnakes.Runtime.CPython.CPythonAPI.PyObject_GetAttr(PyObject ob, IntPtr attr) in C:\workspace\CSnakes\src\CSnakes.Runtime\obj\Debug\net9.0\Microsoft.Interop.LibraryImportGenerator\Microsoft.Interop.LibraryImportGenerator\LibraryImports.g.cs:line 1692
   at CSnakes.Runtime.CPython.CPythonAPI.GetAttr(PyObject ob, String name) in C:\workspace\CSnakes\src\CSnakes.Runtime\CPython\Object.cs:line 98
   at CSnakes.Runtime.Python.PyObject.GetAttr(String name) in C:\workspace\CSnakes\src\CSnakes.Runtime\Python\PyObject.cs:line 135
   at CSnakes.Runtime.CPython.EventLoop.CoroutineTask.Cancel(CancellationToken cancellationToken) in C:\workspace\CSnakes\src\CSnakes.Runtime\CPython\EventLoop.cs:line 85
   at CSnakes.Runtime.CPython.EventLoop.RunForever() in C:\workspace\CSnakes\src\CSnakes.Runtime\CPython\EventLoop.cs:line 297
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
```

We should unregister the scheduling of cancellation requests once a task is done. Also, we should probably bail out early during cancellation if the task has already been completed.

## Reproduction steps

Add the following integration test to `CoroutineTests.cs`

```cs
[Fact]
public async Task MultipleCoroutinesCancellation()
{
    var mod = Env.TestCoroutines();

    using CancellationTokenSource cts = new();

    _ = await mod.TestCoroutine(cancellationToken: cts.Token);

    cts.Cancel();

    _ = await Assert.ThrowsAsync<OperationCanceledException>(() => mod.TestCoroutine(cancellationToken: cts.Token));
}
```

bastbu:
Closes #776

This PR addresses the issue that cancelling a task that already completed would cause the event loop to throw an unhandled exception (and subsequently break the event loop). The main change that addresses the issue is that we check if the canceled task is scheduled before we attempt any cancellation.

Cancellation handling is now unregistered once a coroutine completes, which is an optimization to avoid unnecessary work.

## Missing test coverage

The following test coverage is missing, as it is either an optimization or was hard to provoke in a test:

- Disposal of temporary resources in case of an exception (i.e. disposal of `coroTask` and `cancellationRegistration` to clean up dangling resources.
- Invoking `cancellationRegistration.Dispose();` is an optimization that does not change the correctness behavior (which is already guaranteed by cancelling only tasks that are scheduled, hence removing this logic will not make a test break as the behavior is still correct).
- The PR does not add additional coverage that in-flight tasks are cancelled on event loop disposal.

### DIFF
--- a/src/CSnakes.Runtime/CPython/EventLoop.cs
+++ b/src/CSnakes.Runtime/CPython/EventLoop.cs
@@ -65,8 +65,8 @@ internal sealed class EventLoop : IDisposable
         private PyObject? doneMethod;
         private CancellationToken cancellationToken;
 
-        public enum LifecycleEvent { CancellationRequested, Completed }
-        public event Action<CoroutineTask, LifecycleEvent>? LifecycleChanged;
+        public enum LifecycleChangeKind { Canceling, Completed }
+        public event Action<CoroutineTask, LifecycleChangeKind>? LifecycleChange;
 
         public TaskCompletionSource<PyObject> CompletionSource { get; } = new(TaskCreationOptions.RunContinuationsAsynchronously);
 
@@ -90,7 +90,7 @@ internal sealed class EventLoop : IDisposable
             if (HasCompleted)
                 return;
 
-            LifecycleChanged?.Invoke(this, LifecycleEvent.CancellationRequested);
+            LifecycleChange?.Invoke(this, LifecycleChangeKind.Canceling);
 
             using var cancelMethod = pyTask.GetAttr("cancel");
             cancelMethod.Call().Dispose();
@@ -102,7 +102,7 @@ internal sealed class EventLoop : IDisposable
             if (ProcessConclusion() == TaskStatus.Running)
                 return false;
 
-            LifecycleChanged?.Invoke(this, LifecycleEvent.Completed);
+            LifecycleChange?.Invoke(this, LifecycleChangeKind.Completed);
 
             Dispose();
             return true;
@@ -283,9 +283,9 @@ internal sealed class EventLoop : IDisposable
                                         () => Enqueue(new CancelRequest(coroTask, cancellationToken)),
                                         useSynchronizationContext: false);
 
-                                    coroTask.LifecycleChanged += (_, state) =>
+                                    coroTask.LifecycleChange += (_, kind) =>
                                     {
-                                        if (state == CoroutineTask.LifecycleEvent.CancellationRequested || state == CoroutineTask.LifecycleEvent.Completed)
+                                        if (kind is CoroutineTask.LifecycleChangeKind.Canceling or CoroutineTask.LifecycleChangeKind.Completed)
                                             cancellationRegistration.Dispose();
                                     };
                                 }


src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -255,6 +267,7 @@ private void RunForever()
 
                             var pyTask = this.methods.CreateTask.Call(request.Coroutine);
                             IDisposable? disposable = pyTask;
```
atifaziz:
```suggestion
                            IDisposable disposable = pyTask;
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -285,9 +300,10 @@ private void RunForever()
...
 
                                 request.CompletionSource.SetException(ex);
-                            }
-                            finally
-                            {
+
+                                // Clean up any resources allocated for the task.
+
+                                cancellationRegistration.Dispose();
                                 disposable?.Dispose();
```
atifaziz:
```suggestion
                                disposable.Dispose();
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -52,6 +52,28 @@ private sealed class CancelRequest(CoroutineTask coroTask, CancellationToken can
         public CancellationToken CancellationToken { get; } = cancellationToken;
     }
 
+    private sealed class ScheduledTask(CoroutineTask task, CancellationTokenRegistration cancellationRegistration) : IDisposable
+    {
+        public CoroutineTask Task { get; } = task;
+
+        public void Cancel(CancellationToken cancellationToken) => Task.Cancel(cancellationToken);
```
atifaziz: Why not dispose the registration here so it's done as early as possible?

```suggestion
        public void Cancel(CancellationToken cancellationToken)
        {
            cancellationRegistration.Dispose();
            Task.Cancel(cancellationToken);
        }
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -285,32 +311,34 @@ private void RunForever()
...
+                            scheduled?.Cancel(request.CancellationToken);
                             break;
                         }
                         case (StopRequest, RunState.Running):
                         {
                             state = RunState.Stopping;
-                            foreach (var task in coroTasks)
-                                task.Cancel();
+                            foreach (var task in scheduledTasks)
+                                task.Cancel(default);
```
atifaziz: nit: Make it more explicit:

```suggestion
                                task.Cancel(CancellationToken.None);
```

src/Integration.Tests/CoroutineTests.cs
```diff
@@ -29,6 +29,65 @@ public async Task MultipleCoroutineCalls()
...
+
+        await Assert.ThrowsAsync<TaskCanceledException>(() => RunAsync());
+    }
+
+    [Fact]
+    public async Task CoroutineThatSelfCancels()
+    {
+        var mod = Env.TestCoroutines();
+        var task = mod.TestCoroutineSelfCanceling(TestContext.Current.CancellationToken);
+        await Assert.ThrowsAsync<TaskCanceledException>(() => task);
```
atifaziz: Two things:

- Why not use a simple local function to be the `Act` here and elsewhere for less task/lambda ceremony?
- I would make the strong assertion that the cancellation wasn't due to some `CancellationToken` so chances are good that the source was something else (Python code in this case).

```suggestion
        Task Act() => mod.TestCoroutineSelfCanceling(TestContext.Current.CancellationToken);
        var ex = await Assert.ThrowsAsync<TaskCanceledException>(Act);
        Assert.Equal(CancellationToken.None, ex.CancellationToken);
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -266,15 +279,21 @@ private void RunForever()
...
                                 {
-                                    _ = cancellationToken.Register(() =>
-                                        Enqueue(new CancelRequest(coroTask, cancellationToken)));
+                                    cancellationRegistration = cancellationToken.Register(
+                                        () => Enqueue(new CancelRequest(coroTask, cancellationToken)),
+                                        useSynchronizationContext: false);
+
+                                    coroTask.LifecycleChanged += (_, state) =>
+                                    {
+                                        if (state == CoroutineTask.LifecycleEvent.CancellationRequested || state == CoroutineTask.LifecycleEvent.Completed)
```
atifaziz: nit:

```suggestion
                                        if (state is CoroutineTask.LifecycleEvent.CancellationRequested or CoroutineTask.LifecycleEvent.Completed)
```

src/CSnakes.Runtime/CPython/EventLoop.cs
```diff
@@ -65,6 +65,9 @@ private sealed class CoroutineTask(PyObject pyTask) : IDisposable
         private PyObject? doneMethod;
         private CancellationToken cancellationToken;
 
+        public enum LifecycleEvent { CancellationRequested, Completed }
+        public event Action<CoroutineTask, LifecycleEvent>? LifecycleChanged;
```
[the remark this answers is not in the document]
bastbu: Thanks for the context, applied this with 179b09189b9a39a32f169f55c400dae20f141fe5.

src/CSnakes.Runtime/CPython/EventLoop.cs
bastbu: Agreed - applied both suggestions with 179b09189b9a39a32f169f55c400dae20f141fe5.

